PR #7, opened 2022-01-03
Add gnome-screensaver as a possible compilation option. Closes #6

### LINKED ISSUE #6: Unable to build project on Debian
## Description
The git project at the moment cannot be built on Debian (or Ubuntu) due to the dependency on budgie-screensaver.

As is - budgie-screensaver is not acceptable for inclusion into their repos due to perceived 'vulnerabilities' of dbus-glib-1 (https://github.com/getsolus/budgie-screensaver/issues/8)

Whether we need to worry  about the current budgie-screensaver is another story - probably is part of the budgie 11 roadmap here.  Mate and XFCE screensavers have exactly the same issue but these are legacy and already in the repos - new software like budgie-screensaver is subject to this ruling.

Request here is to add a meson-option (default is budgie-screensaver) and requisite #ifdef's to allow building against gnome-screensaver

If agreed - I'll do the tweaks.  

Note - This request is to remove the technical debt debian currently carries in its packaging - should allow other debian devs to come onboard without hitting this first roadblock.

## Budgie version
10.5.3

## Operating system
debian sid (3 Jan 2022)

meson --prefix=/usr --libdir=/usr/lib

## Actual result
Stops at meson.build assert.

## Expected result
build successful.

fossfreedom:
## Description
Some distros such as Debian do not include budgie-screensaver
in their archive.
This commit provides an optional compilation method to allow
gnome-screensaver to be used for screenlocking noting that
budgie-screensaver is the recommended screenlocker.

### Submitter Checklist


JoshStrobl:
LGTM, thanks!

PR #11, opened 2022-01-04
Prevent crash when deleting twice the same applet

guillotjulien:
## Description

Reported by @fossfreedom on the old repository (https://github.com/solus-project/budgie-desktop/issues/2189). This simply reset the selection state and disable the controls after deleting.

### Submitter Checklist


JoshStrobl:
Thanks!

PR #15, opened 2022-01-04
meson: Drop spurious 3rd positional argument to gnome.compile_resources

jon-turney:
This may become an error with some future meson version:

 src/raven/meson.build:6:0: ERROR: gnome.compile_resources takes exactly 2 arguments, but got 3.

See https://mesonbuild.com/Gnome-module.html#gnomecompile_resources


jon-turney:
See also https://github.com/solus-project/budgie-desktop/pull/2185

JoshStrobl:
Thanks!

PR #52, opened 2022-02-16
Fix some applications showing their own notification popups when paused

EbonJaeger:
## Description
Before this commit, Firefox would show its own notification popups whenever Budgie notifications or popups were paused or disabled. This seems to fix that issue: when notifications are paused, no popups of any kind are shown.

I have no idea why this matters to Firefox, nor how it even detects it, but so far it seems to be working.

This also matches what the Budgie notification server used to do before the rewrite. Previously the expiry variable was set to 0 and all was right with the world.

### Submitter Checklist

### DIFF
--- a/src/daemon/notifications/dbus.vala
+++ b/src/daemon/notifications/dbus.vala
@@ -198,22 +198,6 @@
 			var id = (replaces_id != 0 ? replaces_id : ++notif_id);
 			var notification = new Notification(id, app_name, app_icon, summary, body, actions, hints, expire_timeout);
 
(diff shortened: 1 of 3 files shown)


EbonJaeger:
The issue appears to still not be fixed as it first seemed. It took some time, but it started happening again. The only thing I can think of that changed was the computer going to sleep and resuming.

I'll leave it up to Josh if we want to keep this change, or abandon it.

JoshStrobl:
Well, I think the changes are sensible regardless of whether or not it fully fixes the issue.

PR #54, opened 2022-02-17
Add support for mutter 10 / GNOME42

fossfreedom:
## Description
This add mutter 10 / GNOME 42 support.

This also tidies up and removed mutter 8 / GNOME 40 support

Marked as WIP.

### Submitter Checklist


rodrigofarias77:
Thank you! Fedora Rawhide is happy with this.

fossfreedom:
@rodrigofarias77 thx.  Please do help out testing here - I know various bits like gsettings keys have been removed - so it would be more than helpful if you find any unexpected issues please let me know.

JoshStrobl:
LGTM, feel free to mark as ready for review when you are comfortable getting it merged. Going to see what I can do to test under 42.x on Arch, or maybe through an Ubuntu Daily on a machine.

fossfreedom:
I've tested this on debian sid + its experimental repo that has mutter-10 etc.

Seems to work here.  With the confirmation of working on rawhide, I suppose final confirmation with arch would be nice but I don't know if arch has all the necessary beta packages right at this moment.

PR #57, opened 2022-02-21
Fix notification mute button sometimes starting without an image

EbonJaeger:
## Description
Since the notifications rewrite, the button to trigger Do Not Disturb would sometimes not be shown.

This ensures that it always starts with an icon, and is thus shown.

### Submitter Checklist


JoshStrobl (approved):
Thanks!

PR #59, opened 2022-02-21
Actually fix notification pausing sometimes breaking Firefox notifica…

EbonJaeger:
This time without me messing up entirely with Git.

## Description
This fixes (for real this time) notifications from Firefox not using the system notification implementation when notifications are paused. The issue occurred because per the spec, an empty DBus error with no message should be returned when an application attempts to close a notification, but no notification can be found. Firefox tries to close the notifications it generates, and because there are no notification popups stored, we would send the DBus error. This causes Firefox to stop using the system notification system for the rest of the browser session.

Before the rewrite, Budgie did not send the DBus error, and gnome-shell currently doesn't, either. So, just return instead of sending an error and potentially breaking notifications.

Tested with multiple long periods of having notifications paused over the last day and a half, and Firefox notifications continuing to function as expected.

Thanks to the folks at the Firefox developers Matrix for assisting in finding the cause and helping me look at the Firefox code around notifications!

### Submitter Checklist

### DIFF
--- a/src/daemon/notifications/dbus.vala
+++ b/src/daemon/notifications/dbus.vala
@@ -220,9 +220,6 @@
 							!this.dispatcher.notifications_paused;
 
 			// Set to expire immediately if a popup shouldn't be shown.
-			// This prevents some applications, e.g. Firefox, from showing
-			// its own notification when Budgie notifications are paused for
-			// any reason.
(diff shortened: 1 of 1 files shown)


JoshStrobl (approved):
Thanks!

Issue #56, opened 2022-02-21
Switch Input Source Alt+Shift_L crashes budgie-wm when not using ibus


fossfreedom:
## Description
Raised from Debian bug report: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=999609

## Budgie version
10.5.3

## Operating system
Debian Testing/Ubuntu 21.10

## Steps to reproduce the issue

Dear Maintainer,

   * What led up to the situation?

     Typing specific combination key causes budgie-wm segfault.
     At most twice or so is enough to reproduce this issue.

   * What exactly did you do (or not do) that was effective (or
     ineffective)?

     Typing Alt+Shift_L twice or more.
     Digging a bit, it seems that changing input source is related
     and there are some prerequisite conditions.

      * Step1: Using non ibus as input method
        * For example, set fcitx5 by im-config (Using ibus-mozc)
      * Step2: budgie-desktop keboard layout applet does not show any
          keyboard layouts.
      * Step3: Type Alt+Shift_L some times (twice or so)
           It kills budgie-wm. Note that Shift_L+Alt does not cause it.

     I've found two workarounds:

      * Workaround1: As noted above, use Shift_L+Alt.

      * Workaround2: Launch dconf-editor and remove Alt+Shift_L from
   /com/solus-project/budgie-wm/switch-input-source property.
  
      Before: ['<Alt>Shift_L', '<Super>space', 'XF86Keyboard']
      After: ['<Super>space', 'XF86Keyboard']

* note from 19th Feb 2022

As for 10.5.3+git20220217a-1, it seems that Workaround 2 is not effective anymore.

## Actual result

* What was the outcome of this action?

     It kills budgie-wm process, so desktop session is lost.
     Need to login again.

## Expected result

  * What outcome did you expect instead?

     Alt+Shift_L does not kill budgie-wm process.

## Additional information

Debian Release: bookworm/sid
  APT prefers unstable
  APT policy: (500, 'unstable')
Architecture: amd64 (x86_64)

Kernel: Linux 5.14.0-4-amd64 (SMP w/4 CPU threads)
Locale: LANG=ja_JP.UTF-8, LC_CTYPE=ja_JP.UTF-8 (charmap=UTF-8), LANGUAGE=en_US:en
Shell: /bin/sh linked to /usr/bin/dash
Init: systemd (via /run/systemd/system)
LSM: AppArmor: enabled

Versions of packages budgie-desktop depends on:
[...]

PR #61, opened 2022-02-23
Fix SEGV when input source was switched by Alt+Shift_L

kenhys:
## Description

Closes: #56

When budgie-wm is used with non ibus input method,
It causes a SEGV unexpectedly.

There are two vectors to cause this situation.

* Array<InputSource> is not initialized explicitly, but it may be
  accessed unexpectedly.
* Array<InputSource> may be empty, thus it causes arithmetic exception
 with "% sources.length"

## Steps to reproduce

* Use non ibus input method (e.g. fcitx5)
* Type Alt+Shift_L twice or so

PR fixes the issue on Debian unstable with budgie-desktop_10.5.3+git20220217a-1 (with this patch)

### Submitter Checklist

### DIFF
--- a/src/wm/keyboard.vala
+++ b/src/wm/keyboard.vala
@@ -61,7 +61,7 @@ namespace Budgie {
 		string[] options = {};
 
 		Settings? settings = null;
-		Array<InputSource> sources;
+		Array<InputSource> sources = null;
 		InputSource fallback;
 
 		uint current_source = 0;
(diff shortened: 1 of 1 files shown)


src/wm/keyboard.vala
```diff
@@ -143,6 +149,7 @@ namespace Budgie {
 
 		/* Reset InputSource list and produce something consumable by xkb */
 		void update_sources() {
+			sources = null;
```
fossfreedom, line 152: This looks superfluous - the same variable is immediately initialised afterwards

fossfreedom (changes requested):

src/wm/keyboard.vala
JoshStrobl: The latter shouldn't be required either since it is being created on ln 64 now.

JoshStrobl (changes requested):

src/wm/keyboard.vala
kenhys: I've fixed it in another way.

JoshStrobl (approved):
LGTM, thanks!

PR #62, opened 2022-02-24
Update tab switcher style for built-in theme

serebit:
## Description

Restyles the tab switcher (alt-tab) in the built-in theme to fit with the rest of the dialogs.

### Before

image

### After

image


JoshStrobl (approved):
LGTM, thanks!

PR #63, opened 2022-02-25
Implement fractional scaling for icon-tasklist icons

serebit:
## Description

This commit allows icons in Icon Tasklist to scale to any arbitrary panel size. The scaling factor implemented is 2/3 of the panel size, which gives 24px icons at a 36px panel size (the same as the current default).

### Submitter Checklist


serebit:
This does not implement scaling for Pixbuf icons. Going to close for now, will reopen when I've added those commits.

PR #64, opened 2022-02-26
Proper icon tasklist scaling

serebit:
## Description

This pull request moves icon scaling within the icon tasklist from the applet itself to each of the buttons, and provides a smooth scaling gradient for those icons between different panel sizes. Icon updates are now done on button size allocation, and only if the new allocation differs from the old allocation.

### Old Behavior

Icons would resize in predetermined steps, such as 24px to 48px to 96px.

https://user-images.githubusercontent.com/12981608/157320723-7db1d048-106b-457c-b035-1e92e4bf5fae.mp4

### New Behavior

Icons sourced by gicons are now scaled to a target size that approximates 2/3 of the size of the panel itself, up to an icon size of 54px, where it shifts to the panel size minus 18. This allows for icons to scale to much smaller and much larger sizes, while retaining desirable icon sizes for dock setups, and while maintaining the same icon size for the default panel size of 36px.

https://user-images.githubusercontent.com/12981608/157320237-3efd6160-654d-4bee-901b-5b33371443ea.mp4

**This video is representative of the scaling method, and NOT representative of the target icon sizes produced by the mentioned formulas.**

### Submitter Checklist

### DIFF
--- a/src/applets/icon-tasklist/IconButton.vala
+++ b/src/applets/icon-tasklist/IconButton.vala
@@ -38,6 +38,7 @@ public class IconButton : Gtk.ToggleButton {
 	private Gdk.AppLaunchContext launch_context;
 	private int64 last_scroll_time = 0;
 	private bool needs_attention = false;
+	private int target_icon_size;
 
 	public unowned Budgie.Abomination.Abomination? abomination { public set; public get; default = null; }
 	public unowned Budgie.AppSystem? app_system { public set; public get; default = null; }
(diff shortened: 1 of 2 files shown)


serebit:
Okay, with that last commit I believe this is ready to go. Paging @JoshStrobl for review.

EbonJaeger (commented):
I'm getting a
`(budgie-panel:3746): GdkPixbuf-CRITICAL **: 10:59:48.487: gdk_pixbuf_scale_simple: assertion 'dest_width > 0' failed`
whenever I close Budgie Desktop Settings, even without changing anything. Still not sure if it's caused by this, but maybe worth looking in to. I can only get it to happen with BDS, but it does happen every time.

Other than that, this looks sane to me.

EbonJaeger (approved):
This does solve the Critical message, and everything appears to function properly.

LGTM!

JoshStrobl:
LGTM as well. All the bugs raised were addressed. Great work!

JoshStrobl (approved):

Murbert:
I am new to Github and still figuring my way around so please indulge me.
If I read this correctly, icon scaling is fixed and will show up in Budgie 10.7, is this correct? It's not working for me in 10.6.1.
Thanks

PR #68, opened 2022-03-08
Fix inconsistency when opening Raven

### LINKED ISSUE #31: Raven notifications/applets mismatch
When opening Raven for the first time in a session, it initially shows that the "Applets" section is opened, but what actually appears is "Notifications". Selecting the correct option seems to fix the bug for the remainder of the session, and it defaults back to opening the "notifications" screen from then on.
Screenshot from 2022-01-09 15-11-10

EbonJaeger:
## Description
Per the answer [here](https://stackoverflow.com/questions/22178524/gtk-named-stack-childs), if a stack child is not visible, asking the stack to set the visible child will silently fail, apparently even if the child being shown *is* visible.

The NotificationView widget already calls `show_all()` in its constructor, so the call only had to be added to the applet window. This fixes the case where clicking the notification bell fails to open the correct page, and possibly also fixes where clicking the Raven trigger for the first time sometimes fails to open Raven (it is technically open, but the elements aren't visible).

Thanks to @fossfreedom for finding that link!

Fixes #31

### Submitter Checklist

### DIFF
--- a/src/raven/main_view.vala
+++ b/src/raven/main_view.vala
@@ -77,6 +77,10 @@ namespace Budgie {
 			mpris = new MprisWidget();
 			box.pack_start(mpris, false, false, 0);
 
+			// Make sure everything is shown. Not having this can cause
+			// silent failures when switching stack pages or opening Raven.
+			scroll.show_all();
+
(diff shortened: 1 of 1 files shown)


JoshStrobl:
Thanks!

PR #70, opened 2022-03-15
Improve racyness when a notification is closed and another one shown in quick succession

EbonJaeger:
## Description
There is a race condition where a notification may be closed and in the process of destruction, and a new notification is requested to be shown right after. This causes the position calculation to fail because the existing popup is no longer a Widget, but it's also not `null`.

This problem is addressed here by adding a boolean property to the popup widget that is set to `true` when `dismiss()` is called on it, and checking that property when we want to position a new notification. Additionally, move the latest notification check out of the `switch` branches, and getting our needed values as soon as possible.

### Submitter Checklist

### DIFF
--- a/src/daemon/notifications/dbus.vala
+++ b/src/daemon/notifications/dbus.vala
@@ -291,8 +291,7 @@
 			Gdk.Rectangle rect = mon.get_geometry();
 
 			/* Set the x, y position of the notification */
-			int x;
-			int y;
+			int x = 0, y = 0;
 			calculate_position(popup, rect, out x, out y);
 			popup.move(x, y);
 		}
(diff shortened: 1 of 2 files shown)


serebit:
Fixes my issue with Lollypop, LGTM.

serebit (approved):

src/daemon/notifications/dbus.vala
```diff
@@ -304,27 +303,29 @@
 		private void calculate_position(Popup window, Gdk.Rectangle rect, out int x, out int y) {
 			var pos = (NotificationPosition) this.panel_settings.get_enum("notification-position");
 			var latest = this.popups.get(this.latest_popup_id);
+			bool latest_exists = false;
+			int existing_height = 0, existing_x = 0, existing_y = 0;
+
+			if (latest != null && !latest.destroying) {
```
JoshStrobl, line 309: Could do instead:

```
bool latest_exists = (latest != null && !latest.destroying);
int existing_height = 0, existing_x = 0, existing_y = 0;

if (latest_exists) {
[...]
}
```

JoshStrobl (changes requested):
LGTM otherwise. Pretty minor thing (the change requested not the bug itself :D)

src/daemon/notifications/dbus.vala
EbonJaeger: Good idea. I'll change that sometime this afternoon 😃

JoshStrobl (approved):

PR #86, opened 2022-03-27
Update descriptions for built-in applets

serebit:
## Description
Instead of just duplicated titles, actually describe what each applet does.

### Submitter Checklist


JoshStrobl (approved):
LGTM, thanks!

PR #90, opened 2022-03-29
Add more panel config options to panel.ini

### LINKED ISSUE #88: Add dock ability for panel.ini/layouts
### Current Budgie version
10.6

### Use-cases
Ubuntu Budgie 22.10 is considering dropping Plank and using the inbuilt budgie dock capability now that the icon sizing issues for the panel have been resolved (yay!).  This will mean our default layout will be a true budgie only experience without needing a third-party app.

To enable this will require the ability to specify a panel in dock mode as part of the initial user experience.

### Proposal
At the moment in a panel.ini configuration we can specify extra panels, the applets contained therein, the size and the location of the panel.

The proposal here is to add a series of options in addition to "Position" and "Size" :

"AutoHide" with the values defined in the budgie desktop settings panel settings Automatically Hide drop-down
"Transparency"  with the values defined in the budgie desktop settings panel settings Transparency Hide drop-down
"DockMode" with a True/False value

Maybe needed is "Shadow" and "StyliseRegion" boolean values

### References
https://github.com/BuddiesOfBudgie/budgie-desktop/blob/master/src/panel/manager.vala#L1243

serebit:
Includes autohide policy (Autohide: enum), transparency (Transparency: enum), dock mode (Dock: boolean), and shadow visibility (Shadow: boolean).

Additionally, this PR removes the ability to set panel transparency via a gsettings override. Downstreams that use this method to change default panel transparency will need to ship a modified panel.ini instead.  

Resolves #88.

### Submitter Checklist

### DIFF
--- a/src/panel/manager.vala
+++ b/src/panel/manager.vala
@@ -886,6 +886,8 @@ namespace Budgie {
 			PanelPosition position;
 			PanelTransparency transparency;
 			AutohidePolicy policy;
+			bool dock_mode;
+			bool shadow_visible;
 			int size;
 
 			var settings = new Settings.with_path(Budgie.TOPLEVEL_SCHEMA, path);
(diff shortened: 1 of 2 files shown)


fossfreedom (approved):
Tested this on UB trying to mimic UBs current layout - dynamic top base, dockmode full transparency, sizing etc.

Happy with this.  Testing though does show the current weaknesses of budgie's dock mode implementation.  That's another series of issues.

JoshStrobl (approved):
LGTM as well!

PR #92, opened 2022-03-29
Update timezone if no custom timezone set

Staudey:
## Description
This makes sure that the clock applet refreshes its timezone every update (unless a custom timezone is set), so system timezone changes are reflected without a relaunch of the panel.

This is just a quick and dirty fix (starting point) before I go to bed. There are most likely much cleaner solutions that don't have to run every update cycle (catch systemd-timedated signals?).
Comments very welcome to help me learn how to possibly handle this better :smiley: 

### Submitter Checklist


JoshStrobl (approved):
Seems fine to me. @serebit, @EbonJaeger, @guillotjulien have any other thoughts or ideas?

serebit:
This works, but I'm not super keen on creating a new TZ object every second. As per the discussion on Matrix, there appears to be a [PropertyChanged signal](https://www.freedesktop.org/software/systemd/man/org.freedesktop.timedate1.html#Properties) in systemd that can be subscribed to for monitoring time zone changes.

serebit:
I've decided that this is a reasonable change, and I'd rather keep things portable than rely on systemd for time zone updates. Therefore, LGTM and thanks!

Issue #93, opened 2022-03-30
Multiple apps in ITL shown as "active"


serebit:
## Description
In some cases, Budgie's icon tasklist will show multiple apps as "active", even if one of those apps is not in focus. It seems as though the "active" state can get stuck on some apps.

![image_20220330102743](https://user-images.githubusercontent.com/12981608/160869701-845cfee6-8597-468c-a347-a58cb6804aba.png)

Steam, Discord, and Kitty are not pinned.

## Budgie version

master @ be79b9c

## Operating system

EndeavourOS

## Steps to reproduce the issue
To reproduce with Discord:

- Close Discord entirely
- Open a different app, like gedit, and focus on it
- Open Discord and wait until the main window appears

Discord should now be stuck as "active" until the main window is closed and reopened.

serebit:
Not fixed as of be79b9c. To reproduce with Discord:

- Close discord entirely
- Reopen it, and while it's loading, before the main window appears, switch focus to a different app

Discord should now be stuck as "active" until the main window is closed and reopened.

serebit:
Reproduced by @Staudey with v10.6.1 on Solus. Unfortunately, both Guillot and Josh are unable to reproduce it, so it's on me to attempt to fix it.

serebit:
Bisected - offending commit is 07007bf.

PR #94, opened 2022-03-30
Fix regression where non pinned icons were not reorderable in icon tasklist

guillotjulien:
## Description

This was due to the incorrect app ID being passed to the signal handler.
Instead we can pass the correct ID that we determine in IconTasklistApplet when instanciating the button.

This revert last attempt to fix the issue as it was causing other issues with Budgie settings and LibreOffice (and quite possibly the issue reported by @serebit in #93)

### Submitter Checklist

### DIFF
--- a/src/applets/icon-tasklist/DesktopHelper.vala
+++ b/src/applets/icon-tasklist/DesktopHelper.vala
@@ -75,17 +75,17 @@ public class DesktopHelper : GLib.Object {
 	}
 
 	/**
-	* get_app_launcher will return the last past of an app_id string. Useful when handling the full path to a DesktopAppInfo
-	*/
-	public string get_app_launcher(string app_id) {
-		string[] parts = app_id.split("/");
-		return parts[parts.length - 1];
-	}
-
-	/**
 	 * Return the currently active workspace
 	 */
 	public Wnck.Workspace get_active_workspace() {
 		return screen.get_active_workspace();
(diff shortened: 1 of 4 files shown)


JoshStrobl:
No. Base it off the work I did.

JoshStrobl:
Re-opening since I will be nuking non-grouping anyways.

src/applets/icon-tasklist/DesktopHelper.vala
```diff
@@ -74,14 +74,6 @@ public class DesktopHelper : GLib.Object {
...
 	}
 
-	/**
-	* get_app_launcher will return the last past of an app_id string. Useful when handling the full path to a DesktopAppInfo
-	*/
-	public string get_app_launcher(string app_id) {
-		string[] parts = app_id.split("/");
-		return parts[parts.length - 1];
-	}
-
```
JoshStrobl, line 84: Would still like this moved to DesktopHelper in case we need to use it elsewhere.

JoshStrobl (commented):
Seems fine otherwise.

src/applets/icon-tasklist/DesktopHelper.vala
guillotjulien: Sure

JoshStrobl (approved):
Thanks!

PR #97, opened 2022-04-01
Initial multipanel implementation

fossfreedom:
## Description
This is for iterating on only - i.e. not ready in any shape or form - feel free to take run/hide/reimplement/whatever. 

The purpose of this is to show the areas of change if we want to implement multiple panels across multiple monitors

Current implementation has significant issues - so be aware how you adjust your monitor setup in budgie-control-center

1. we cannot have two panels in the same position i.e. two top panels on different monitors
2. the set_struts routine needs reworking to understand that panels can exist on different monitors and struts need to be per monitor
3. disconnecting a monitor should handle moving panels/not moving panels from the monitor to the primary display - it doesnt at the moment even though I think I have connected to the right display signals

So what this test code shows is that:

1. when no monitors are connected (i.e. just a laptop screen) budgie-desktop-settings panel settings should just look normal
2. when a monitor is attached (logout and login I think) - panel settings should have a new option to set the display for the panel - the dropdown will have the monitor names
4. set - for example, one top panel on the laptop screen, set a second panel - right panel on the monitor screen.
5. add applets on each of the panels
6. logout and login - panels should be displayed as you left them - i.e. laptop top panel, monitor right panel.


JoshStrobl:
@fossfreedom can we close this and just tackle it in 11?

PR #99, opened 2022-04-05
Per-panel setting for applet spacing

serebit:
## Description

This pull request adds a per-panel setting that defines the base spacing between applets, independent of CSS. This works via setting the spacing of each individual panel region, in addition to the spacing between the regions, so spacing between applets is consistent even at region boundaries. The default setting of 2 matches our current no-setting default, and the setting can be adjusted from a minimum value of 0 to a maximum value of 64.

The spacing is also adjustable in `panel.ini` via the `Spacing` key.

### Submitter Checklist


JoshStrobl:
LGTM, thanks!

PR #100, opened 2022-04-06
Use meson warning_level instead of global warning flags

serebit:
## Description

This PR is a simple change that sets `warning_level` to 3 across the entire project, with an override for the gvc subproject to a `warning_level` of 0 (as we can't fix any of these warnings due to it being a subproject, there's no real point in printing them). Additionally, we now set the project_name and project_version args when compiling gvc.This results in a reduction in compile log lines from 879 to 546 on current master, and additionally, still prints every warning in the main Budgie codebase that was printed with the global cflags.

### Submitter Checklist

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -8,6 +8,7 @@ project(
     ],
     default_options: [
         'c_std=c11',
+        'warning_level=3'
     ],
 )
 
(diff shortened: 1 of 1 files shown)


JoshStrobl:
Thanks!

PR #102, opened 2022-04-07
Use a Mutex for Raven threading

serebit:
## Description

Uses a Mutex for synchronization of prop and anchor changes. This SHOULD have the same result as the prior code, but without using deprecated functions to force synchronous behavior.

### Submitter Checklist


JoshStrobl:
Yea...that is way better.

Issue #24, opened 2022-01-06
Improving Budgie Menu


JoshStrobl:
This is a meta task to highlight actionable items that would improve Budgie Menu, make it more flexible for users, and reduce the need for third-party solutions or forks of stuff like Slingshot.

Firstly, the desktop application tracking needs to be improved. While libgmenu / gnome menu has served us well and I'm not really opposed to its continued usage, we do run into circumstances where we unnecessarily rebuild the entire tree before building a new view. This results in a temporary empty view of Budgie and sometimes can be a problem, such as not showing items consistently.

A Budgie Menu update should take into account a desired rearchitecture to only partially manipulations the existing tree, or builds a new tree that contains references to the old one (where unchanged) and new one (new additions, deletions, etc.) and only then pushes it to the menu to be rebuilt.

Alongside this, this centralized "store" in Budgie Menu should be view agnostic. This would enable us to build out alternative views in Budgie Menu, such as an icon grid / icon only view to replace alternatives like Slingshot.

Alongside this, I personally would like to see us default to not having a User Indicator, but rather have that functionality in Budgie Menu as well, which provides familiarity for users of other desktop environments like KDE and MATE + Budgie Menu, as well as Windows users.

grahammtbr:
My biggest need/want with a launcher/start menu is the ability to for it to remember my most used applications, and rank them higher in search. I believe that Budgie Menu had this once upon a time, and until recently Ubuntu Budgies "Slingshot" launcher had this via Zeitgeist (not sure why that one recently disappeared). Maybe I just need to break my old habits of hitting the super key plus a couple characters and blindly hitting enter, but since the removable of this feature from Slingshot I've done nothing but constantly open wrong apps!

And sorry, I'm not a programmer that can help with this unless you want it written in PHP :laughing:

rmrfforewardslashstar:
The functionality of the [GNOME extension "Arc Menu"](https://gitlab.com/arcmenu/ArcMenu) might be something to base customization functionality on or act as inspiration. This is always the first extension I install while using GNOME.

chaibronz:
Just realized that the current menu doesn't respect all panel regions if using multiple panels. It will respect the region of the panel it is on, but not any adjacent. This results in the menu potentially overlapping other panels when open. Any chance this could be looked at in the re-write? 

Expected behavior -- in the screenshot below -- would be for the top of the menu to open below the top panel. 

[...]

PR #110, opened 2022-04-12
Rewrite budgie menu

### LINKED ISSUE #127: Applications not launching on the correct GPU
## Description

I have a dual-GPU optimus laptop. On GNOME, applications are launched on the iGPU (here, intel iRIS Xe) as default and on the dGPU (here, NVIDIA MX450) when the .desktop file includes the line `PrefersNonDefaultGPU=true`. On budgie, everything launches on the iGPU no matter what, leaving me with poor performance in games.

## Budgie version

budgie-desktop 10.6.1

## Operating system

Pop!_OS 22.04

1 : Be in hybrid mode (Pop!_OS has multiple modes, hybrid is the default one with both cards activated)
2 : Launch any application that has the line `PrefersNonDefaultGPU=true` in their .desktop file

## Actual result

Observe (with e.g. `nvtop`) that the application was launched with the wrong GPU

## Expected result
The application should launch on the NVIDIA GPU if it has the corresponding line.

## Additional information
I use proprietary NVIDIA drivers, version 510.68.02.

EbonJaeger:
## Description
This is a large rewrite/refactor of the Budgie Menu applet.

Gone are the days of installing a new application, opening the Budgie Menu to launch it, and not being able to find it. The new and improved Budgie Menu swaps out GMenu for a new tracking system based on GLib's AppInfoMonitor. With the new implementation, tracking installed applications is much more reliable, and applications should be present in the menu as soon as they're installed.

### Dual GPU Support

Budgie Menu now leverages [switcheroo-control](https://gitlab.freedesktop.org/hadess/switcheroo-control) when available to switch between GPU environments when launching applications that request it. Users with dual-GPU setups can now launch their applications without necessarily having to worry about which GPU is in use and manually switching between them.

### Power & Session

Controls for session management can now be found right in the Budgie Menu. You'll now be able to shut down and restart your computer, log out or lock your screen, all from the Budgie Menu without having to look for the `user-indicator` applet on your panels.

https://user-images.githubusercontent.com/5157277/163042930-7b8f0786-21e4-4335-a03d-88ba5d6ac884.mp4

### Abstracted Views

The menu has been re-architectured to allow much easier development of different ways to present applications in the menu. This doesn't mean much right now, but some people prefer having a grid, and this brings us one step closer to enabling just such a view.

### Search Changes

Searching has been changed a bit with the new Budgie Menu. As an end user, the only real change you should see is that searching is now a bit more "fuzzy." This means that if you don't type an application's name exactly, it should still be a result as long as you're close.

For developing, searching and result relevancy is handled by a new central class, making it much easier to follow and adjust as needed later on. This also means that we don't have to calculate relevancy multiple times per application for a single search (e.g., for filtering and sorting).

### Settings

A new setting has been introduced to address the desire to show or hide Budgie Control Center items in the applications menu. By default, these items will not be shown, in order to match the behavior of Budgie 10.6. If users want to have these items shown, they can enable this in the applet settings in Budgie Desktop Settings. When searching, control center items are shown regardless of this setting, so that they are always accessible even when hidden.

### Theming

To make it easier for GTK theme developers to theme the Budgie Menu, several new CSS classes have been added:

- `budgie-menu-header`: Applied to the GtkBox holding the search entry widget
- `budgie-menu-footer`: Applied to the GtkBox holding the user and power buttons
- `budgie-menu-overlay`: Applied to the GtkListBox holding menu items. These widgets (in the event that there are more than one) are a part of a GtkOverlay as overlay widgets
- `user-icon-button`: Applied to the user button in the footer

The `budgie-menu-overlay` one is especially important, as some themes will give the power menu a transparent background without special handling because it's a part of a GtkOverlay.

### Building

This work adds two dependencies to Budgie:

- cairo
	+ Solus: `libcairo-devel`, Ubuntu: `libcairo2-dev`
- gee
	+ Solus: 'libgee-devel', Ubuntu: `libgee-0.8-dev`

### Screenshots & Videos

image

### Submitter Checklist

Ref #24
Fixes #127


src/applets/budgie-menu/backend/Tracker.vala
```diff
@@ -0,0 +1,225 @@
...
+		});
+
+		this.categories.add(new Category(_("Games")) {
+			included_categories = { "Game" }
+		});
+
+		this.categories.add(new Category(_("Graphics")) {
+			included_categories = { "Graphics" },
+			excluded_applications = {
+				"eog.desktop",
```
JoshStrobl, line 123: EOG should not be excluded. It deals with graphics (images) and is currently listed under Graphics.

src/applets/budgie-menu/backend/Tracker.vala
```diff
@@ -0,0 +1,225 @@
...
+				app.get_commandline().length != control_center.length
+			);
+
+			// Check if we should not add this application to the store
+			bool should_skip = (!desktop_app.should_show() && !is_control_center_panel) || // CC items are handled by the UI
+								(desktop_app.get_boolean("Terminal"));
+
+			if (should_skip) {
+				continue;
+			}
```
JoshStrobl, line 200: I can't say I am a fan of this logic being here. The idea of the tracker was solely to provide a central store for appinfo, with unified search logic. If something should be excluded on the presentation layer, like Budgie Menu, it should be done at the presentation layer.

src/applets/budgie-menu/views/ListView.vala
```diff
@@ -0,0 +1,483 @@
...
+			btn.toggled.connect(() => {
+				this.update_category(btn);
+			});
+
+			btn.show_all();
+			this.categories.pack_start(btn, false, false, 0); // Add the button
+
+			// Create a button for each app in this category
+			foreach (var app in category.apps) {
+				var app_btn = new MenuButton(app, category, icon_size);
```
JoshStrobl, line 188: IMO the logic should be here instead of Tracker for determining if we should show the control panels and skip if the app is a panel and the setting isn't enabled.

JoshStrobl (changes requested):
Extremely good work. Haven't had the chance to actually take a look yet otherwise, but provided some initial feedback here.

I would also like to mention that the "tracking" files should be moved out of the applet code. That should be presentation logic and the tracking, relevancy, etc. (data layer) in somewhere like `src/appindexer` or something (maybe a more apt name for it than Tracker? Real bikeshedding there so can't say I actually care. Honestly.).

src/applets/budgie-menu/backend/RelevancyService.vala
```diff
@@ -0,0 +1,196 @@
...
+	/**
+	 * Determine a score in relation to a given search term.
+	 *
+	 * Somewhat unintuitively, the lower the score returned,
+	 * the more relevant this item is. This is because we take
+	 * the Levenshtein Distance between the term and the name
+	 * into account, and it's way easier to have an arbitrary
+	 * threshold when you have a set minimum value.
+	 *
+	 * Totally stole (most of) this from Brisk (Which I (Josh)
```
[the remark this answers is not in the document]
EbonJaeger: Guess I didn't go far enough back in the history for the old comment line xD
Not sure what to do about it now.

src/applets/budgie-menu/backend/Tracker.vala
EbonJaeger: The problem is that the .desktop files for BCC panels currently have `NoDisplay=true` or whatever the syntax is. Should that then be changed and this check here removed?

JoshStrobl: Think I was misunderstanding the purpose of this. Feel free to ignore.

src/applets/budgie-menu/backend/RelevancyService.vala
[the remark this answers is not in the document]
JoshStrobl: Probably just: "This was inspired by prior art from Brisk Menu." ?

JoshStrobl (changes requested):

JoshStrobl:
Why are you requesting my review on this? I reviewed it and there is still a move of the tracker and relevancy bits that need to be moved out of the applet codebade.

serebit:
The button was sitting there demanding that I press it.

JoshStrobl:
@serebit @EbonJaeger Could you please review these changes and be pedantic about my implementation. Thanks!

serebit (changes requested):
Couple items.

1. The built-in theme has not been updated to accommodate the XDG dirs view. I can address this separately, though.
2. If one of the XDG dirs in the list resolves to a symlink to a folder rather than a regular folder, the file manager will still open, but it will open with a view of `~` instead of the target folder.

LGTM otherwise.

src/applets/budgie-menu/OverlayMenus.vala
```diff
@@ -0,0 +1,313 @@
...
+		} catch (Error e) {
+			warning("Unable to connect to Budgie XDGDirTracker: %s", e.message);
+		}
+	}
+
+	private void handle_xdg_dirs_changed(UserDirectory[] dirs) {
+		this.items_to_show = new List<UserDirectory>(); // Remove all items
+
+		for (var i = 0; i < dirs.length; i++) { // For each directory
+			this.items_to_show.append(dirs[i]);
```
EbonJaeger, line 185: Instead of sorting each time, could we do `insert_sorted()`? I suppose it might be a case of six of one, and half a dozen of the other.

EbonJaeger (commented):
This looks fantastic! Really the only thing I have is insert_sorted vs invalidate_sort, which as I said in the comment might not actually have any difference, in which case it doesn't matter.

A couple of the files (I think the two added ones) don't have a newline at the end of the file, do we care about that at all?

Thanks for doing all of this work!

src/applets/budgie-menu/OverlayMenus.vala
JoshStrobl: We aren't sorting the items_to_show as they do not need to be sorted. The items being sorted are folder_items and those do not get sorted during the handling of the incoming XDG dirs list from the DBus Server. What is being called after the new List is being created is the filtering invalidation, since we are re-checking if an item should be shown / hidden (e.g. Desktop folder gets deleted -> hide item, Music folder gets created -> show item).

JoshStrobl:
> The same thing happens if the directory doesn't exist at all.

You mean if the target directory does not exist / symlink is bork?

The XDG directories should be dynamically shown / hidden based on `filter_list_box_item` which checks if our items_to_show List contains the respective XDG dir, if not it gets filtered out from the ListBox view, but indeed if the issue is with symlinks I can see what would be an issue.

JoshStrobl:
> A couple of the files (I think the two added ones) don't have a newline at the end of the file, do we care about that at all?

Overlay Menus shows for me the content ending at 313 and there being a newline at 314. For the new dbus.vala it is showing content ending at 385 and newline at 386 :thinking:

JoshStrobl:
Symlinks should now be fixed.

JoshStrobl:
Overlay menu styling refined.

serebit (approved):

JoshStrobl (approved):

EbonJaeger:
Everything looks good here! Thank you, @JoshStrobl, for your work, @serebit for your styling, and everyone who tried it out, tested it, and offered suggestions and feedback. This is set to be merged, and if anything comes up after, it can be addressed separately.

PR #122, opened 2022-05-13
GitHub Actions workflow for Ubuntu 22.04

serebit:
## Description

Now that Ubuntu 22.04 has been released and GitHub Actions has a beta image for it, Budgie can build from within GitHub Actions. This PR adds a workflow to build Budgie on Ubuntu 22.04 with default options, except for use of GNOME Screensaver.

### Submitter Checklist

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -10,6 +10,6 @@ jobs:
         python-version: '3.x'
     - run: pip install meson ninja
     - run: sudo apt update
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -0,0 +1,11 @@
+name: Build on Ubuntu 22.04
+on: [push]
```
JoshStrobl, line 2: This will apply to every branch, which isn't ideal, especially if those are WIP / drafts. In my opinion, master branch pushing is fine and PRs to either master or the 10.6 release branch.

JoshStrobl (changes requested):

JoshStrobl (approved):
Perfect! Thanks :100:

PR #123, opened 2022-05-14
Rewrite initial applet layout to prevent position desync

### LINKED ISSUE #72: Spacer at the edge of a panel isn't preserved after reboot
## Description
I added a spacer to the right edge of my top panel so that the widget wouldn't be squished into the corner so much. While I am able to manually accomplish the desired behavior with the Budgie Desktop Settings, the spacer isn't shown anymore after a reboot. 

## Budgie version
budgie-desktop 10.5.3 (git-26cde8fd300a87aba0d085ca50c89728150138d2)

## Operating system
Ultramarine Linux 21.12 Revision 1

1. add a spacer to the right (or probably any edge) of your panel
2. reboot

## Actual result
the spacer isn't shown in the panel

## Expected result
some space between the edge of the screen and the rightmost widget

## Additional information
I have a spacer left to the rightmost widget as well to make it stand out more.

serebit:
## Description

Rather than adding applets directly to the panel on initial panel load, the `load_applets` function instead uses intermediate region lists sorted by the info's desired position. Once all applet info have been added to the region lists, each applet's info is updated with its actual position within the list, and is then added to the panel. This prevents issues with applets being uninstalled before being removed from the panel, and thus should resolve #72, along with other applet ordering issues that could be the result of initial race conditions.

### Submitter Checklist

### DIFF
--- a/src/panel/panel.vala
+++ b/src/panel/panel.vala
@@ -518,7 +518,6 @@ namespace Budgie {
 				return (int) (a.position > b.position) - (int) (a.position < b.position);
 			};
 
-			/* Two loops so we can track when we've fully loaded the panel */
 			lock (expected_uuids) {
 				for (int i = 0; i < applets.length; i++) {
 					this.expected_uuids.append(applets[i]);
@@ -533,10 +532,8 @@ namespace Budgie {
 					Budgie.AppletInfo? info = this.manager.load_applet_instance(applets[i], out name);
 
 					if (info == null) {
-						/* Faiiiil */
 						if (name == null) {
 							unowned List<string?> g = expected_uuids.find_custom(applets[i], strcmp);
-							/* TODO: No longer expecting this guy to load */
 							if (g != null) {
 								expected_uuids.remove_link(g);
 							}
@@ -551,7 +548,6 @@ namespace Budgie {
 						}
 					}
 
-					/* um add this bro to the panel :o */
 					if (info.alignment == "start") {
 						start_applets.insert_sorted(info, infocmp);
 					} else if (info.alignment == "center") {
--- a/src/panel/panel.vala
+++ b/src/panel/panel.vala
@@ -514,11 +514,19 @@ namespace Budgie {
 				return;
 			}
 
+			CompareFunc<Budgie.AppletInfo?> infocmp = (a, b) => {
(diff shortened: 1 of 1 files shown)


src/panel/panel.vala
```diff
@@ -533,15 +541,38 @@ namespace Budgie {
...
+
+						info = this.add_pending(applets[i], name);
+						if (info == null) {
 							continue;
 						}
-						this.add_pending(applets[i], name);
-						manager.modprobe(name);
+					}
+
+					/* um add this bro to the panel :o */
```
JoshStrobl, line 554: Probably not necessary.

src/panel/panel.vala
```diff
@@ -533,15 +541,38 @@ namespace Budgie {
...
+					add_applet(start_applets.nth_data(i));
+				}
+				for (int i = 0; i < center_applets.length(); i++) {
+					center_applets.nth_data(i).position = i;
+					add_applet(center_applets.nth_data(i));
+				}
+				for (int i = 0; i < end_applets.length(); i++) {
+					end_applets.nth_data(i).position = i;
+					add_applet(end_applets.nth_data(i));
+				}
```
JoshStrobl, line 575: Worth maybe doing some g_list_free at the end on the start, center; and end applets?

https://valadoc.org/glib-2.0/GLib.g_list_free.html

JoshStrobl (changes requested):
Only the first review item is a blocker. Kinda just pointless gendering of an applet and doesn't really add any value to the documentation.

The later review item I can't say I care much about. Vala memory management is lacking anyways.

src/panel/panel.vala
serebit: For the record, that was already there prior to this PR, I just didn't remove it. I'll delete it though.

JoshStrobl: Ah. My bad.

src/panel/panel.vala
serebit: The lists are freed once the function finishes, so this shouldn't be necessary.

JoshStrobl (approved):
Thanks for the excellent work on this. Validated locally as well and did not spot any regressions. :100:

PR #124, opened 2022-05-16
Introduce our native screenshot capability.

fossfreedom:
## Description
Developed on UB 22.04 by primarily team member Jacob and myself.

On install (logout and login), a new menu item will appear - budgie screenshot.

Fullscreen, per window and area screenshots are possible - with/without cursor and a delay option is available.

Taking a screenshot will present another window where you can save/cancel/copy to clipboard/open in the image editor. Save option includes saving to any folder assuming you have permissions.  This saves as you last-saved folder.

Keyboard controls Print Screen, Ctrl Print Screen and Alt Print screen similarly invokes the three screenshot types.

----

structure:

Everything has to be async to ensure no thread and process blocking.
In the window manager screenshot of the three types are undertaken when a dbus call is made.

For the GUI - this is also a dbus interface that lives within budgie-daemon - the budgie-screenshot application simply invokes the GUI dbus interface.


serebit:
Widget positioning looks off on some of the themes I've tried.
## Matcha

Button icons and text are too far left, and the spinner box and switch are misaligned.

image

## Materia

Button icons and text are too far left, and the spinner box and switch are misaligned (but in reverse).

image

## Plata

Camera icon is too far up, icons and text are too far down.

image

## Arc

Button icons and text are too far left.

image

serebit:
For any default folders that don't actually exist, or are resolved as symbolic links, the directory picker in the after-shot window displays those nonexistent folders with the name of the current user.

fossfreedom:
hmm @serebit obviously there is an issue here ... but I can't reproduce it.

I've tried renaming my Pictures folder and then creating a symlink called Pictures to the renamed folder.  It still appears as Pictures in the drop-down.

I've deleted a folder and it is removed from the drop-down.

So there is something here with your specific setup I haven't yet tried...

src/wm/wm.vala
```diff
@@ -101,6 +104,17 @@ namespace Budgie {
...
+	/**
+	* Allows us to invoke the screenshot client without directly using GTK+ ourselves
+	*/
+	[DBus (name = "org.buddiesofbudgie.ScreenshotControl")]
+    interface ScreenshotControl : GLib.Object {
+        public async abstract void StartMainWindow() throws GLib.Error;
+        public async abstract void StartAreaSelect() throws GLib.Error;
+        public async abstract void StartWindowScreenshot() throws GLib.Error;
+        public async abstract void StartFullScreenshot() throws GLib.Error;
+    }
```
JoshStrobl, line 116: Indentation bork here.

JoshStrobl (changes requested):
Lots of stylistic problems, particularly function arguments having a whitespace between them and the function name, e.g.

```
save_image (...)
```

should just be:

```
save_image(...)
```

Additionally, `else` should be inlined, e.g. `} else {` instead of being on a new line. Catch should be as well.

JoshStrobl:
I tried testing the area-based screenshotting again and it only worked on one of my monitors, which was my left side non-primary monitor. Neither my primary nor my third monitor had the selection area cursor.

xrandr info:

```
Screen 0: minimum 8 x 8, current 7360 x 2160, maximum 32767 x 32767
DVI-D-0 disconnected (normal left inverted right x axis y axis)
HDMI-0 connected 1920x1080+0+0 (normal left inverted right x axis y axis) 477mm x 268mm
HDMI-1 connected 1600x900+5760+0 (normal left inverted right x axis y axis) 443mm x 249mm
DP-0 connected primary 3840x2160+1920+0 (normal left inverted right x axis y axis) 597mm x 336mm
DP-1 disconnected (normal left inverted right x axis y axis)
DP-2 disconnected (normal left inverted right x axis y axis)
DP-3 disconnected (normal left inverted right x axis y axis
```

JoshStrobl:
@fossfreedom I fixed it in 144bbb09d1e711e5ec6f4d51fd22d5d48da13775. It was failing to span across multiple monitors, it was only doing it at the monitor at 0,0 on my configuration.

JoshStrobl:
Did some more changes (see 8b04e0e1b23405f8ba3cc550fab84883609f4e1a). Refinements welcome if someone wishes to raise planned ones, otherwise will merge based on the current implementation.

fossfreedom:
Changes from a quick glance look ok.

plank is now displaying two icons when invoking screenshot from the menu - the wmclass naming change has probably broken this. I'll need to have a look at why before merging.

I'll do some more testing.

JoshStrobl (approved):
LGTM!

Issue #27, opened 2022-01-06
Demystifying Theme Selection


JoshStrobl:
This is a proposal for demystifying theme selection and providing a potentially simpler user experience for theme selection and introducing better cross-toolkit support.

**Current Situation**

At the moment our theme section breaks down to the following:

- A GTK widget theme selection with all items except those in our exclusion list like Adwaita / "Default"
- "Dark theme" toggle which only sometimes does something, depending on the theme or if we are using the built-in theme. This will currently set "gtk-application-prefer-dark-theme". Some themes respect this and others expect you to explicitly choose a different GTK Theme.
- Built-in theming, which is intended to be redesigned #22

**Proposal**

My proposal is to retain both our setting and respecting of `gtk-theme`, however present a UI that translates that option (alongside the dark theme preference). We would do this by us:

1. Maintaining a map of GTK widget themes to dark / light theming (we can also figure out if there is room to have theme developers ship a file that specifies which GTK theme to use in certain scenarios)
2. Leveraging that map to provide a simplified Widgets list. So instead of say "Materia-dark" you would just have "Materia".
3. Translating selections such as "Dark Theme" into "Materia-dark", a "Compact Mode" appearing where we know the theme provides one (so you could imagine it being "Materia-dark-compact" and "Materia-dark-compact" respectively). I'm open to ideas on how to handle scenarios like *just* Plata where it may be light with dark GTK Header, or if that is something we want to concern ourselves with entirely. This would still set the `gtk-theme` in the background, just the Widgets selection itself wouldn't directly.

Built-in theme would still be presented. If enabled, we would use that and choose the respective light / dark theming based on Dark theme being set. If it's off, it'd still use any support from the GTK theme, same situation as now.

Another key difference with the Dark theme toggle is I want us to introduce support for FreeDesktop Dark Style Preference. Toggling it on would set that option and at least ensure applications built with the likes of libadwaita would respect our dark theme user choice.

serebit:
This would be a bit of an undertaking. We'd need to account for accent colors (provided by themes like Matcha), compact versions, hiDPI versions, dark and light and no-pref versions, and whatever other variations theme devs would come up with. The most rigid solution I can think of would be a configuration file provided by GTK themes that provides information about each variant and what properties it has, but adoption would be the main concern there along with potentially shoehorning theme developers into limiting their variants. 

[...]

PR #128, opened 2022-05-23
Add preferred application styling with some logic

fossfreedom:
## Description
This adds the the GNOME 42 color-scheme support at a basic level.  It is intended to be the starter for 10 implementation to the larger discussion piece how to best handle libadwaita stuff.

A new drop down exists in budgie desktop-settings allowing a user to set the color-scheme styling preference.

Also some libhandy type logic has been added to set the color-scheme to dark if the chosen theme has the "-dark" name (and similar logic for light). This is the hard-coded libhandy "themes must end with -dark or -light".  Suggest future PRs will deal with additional logic if deemed necessary.

This removes the "dark-theme" switch toggle - doesnt make sense anymore - at least directly.  The dark-theme is now connected to the styling preference i.e. setting "preferred dark" will set the dark-theme gsettings value to its dark boolean value.

### Submitter Checklist


fossfreedom:
> I would recommend splitting the theme name by dashes, spaces, and underscores, then looking for the word "dark" in the array of words. Some themes don't end with "dark" - Matcha, for example, is named as `Matcha-{dark,light,}-<color>`.

fair enough. done.

JoshStrobl:
> This removes the "dark-theme" switch toggle - doesnt make sense anymore - at least directly. The dark-theme is now connected to the styling preference i.e. setting "preferred dark" will set the dark-theme gsettings value to its dark boolean value.

I am not entirely sure I agree with that and it runs counter to the objectives of #27. Realistically we should swapping both the theme (e.g. Materia -> Materia-dark) and the "standard" preference. Otherwise we are trading one user complexity and inconsistency for another.

EbonJaeger:
Is this pre-empted by https://github.com/BuddiesOfBudgie/budgie-desktop/pull/444?

serebit:
Yeah, I think this has been effectively replaced by #444 and can probably be closed.

PR #130, opened 2022-05-23
Set a mininum expiry timeout for notification popups

### LINKED ISSUE #129: Notifcation popup does not appear for some web browsers
## Description
Notifications for some web browsers do not show a popup notification window

## Budgie version
10.6.1

## Operating system
Ubuntu Budgie 22.04

While using Firefox snap, go to https://pushalert.co/demo and create a demo notification.  No notification pop up appears.  However, the notification does immediately show up in the Raven notification area in the sidebar.

However, while using Firefox ESR, go to the same site and create a demo notification.  Using this browser, the notification popup correctly appears and will disappear and show up in the Raven notification area after the correct decay time.

While I thought this was a snap issue, it seems to affect Chromium based browsers as well, including .deb versions (tested Opera / Vivaldi)

## Additional information
Using dbus-monitor and looking at the difference in the output from Firefox ESR vs Firefox Snap, the only significant difference I could find was:

Firefox ESR:
```
      dict entry(
         string "desktop-entry"
         variant             string "Firefox"
      )
```

Firefox Snap:
```
      dict entry(
         string "desktop-entry"
         variant             string "/snap/firefox/1377"
      )
```

I am unsure of any reason why this would prevent the pop-up, but figured more information is better than less.

EbonJaeger:
## Description
Some applications set a timeout of 0, which the spec says means that it should never expire. Our implementation doesn't really fit that, since persistence is done by Raven, not the notification server. So, we should have a minimum show time for the popups to ensure that one is actually shown when it should be.

Fixes #129

### Submitter Checklist

### DIFF
--- a/src/daemon/notifications/dbus.vala
+++ b/src/daemon/notifications/dbus.vala
@@ -196,7 +196,18 @@
 			int32 expire_timeout
 		) throws DBusError, IOError {
 			var id = (replaces_id != 0 ? replaces_id : ++notif_id);
-			var notification = new Notification(id, app_name, app_icon, summary, body, actions, hints, expire_timeout);
+			var expires = expire_timeout;
+
+			// The spec says that an expiry_timeout of 0 means that the
+			// notification should never expire. That doesn't really make
+			// sense for our implementation however, since this only handles
+			// popups. Notification "storage" is done seperately by Raven.
+			if (expires < 6000) {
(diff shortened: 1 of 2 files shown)


JoshStrobl (changes requested):
Surely all of that could be simplified to being a ternary like with the id?

JoshStrobl:
Perfection.

PR #134, opened 2022-05-28
Korean Translation Update

MarongHappy:
Update Korean Translation

## Description

### Submitter Checklist


JoshStrobl:
This is not the appropriate method. It should be via Transifex. https://www.transifex.com/buddiesofbudgie/budgie-10/

MarongHappy:
OK. Thx.

PR #142, opened 2022-06-12
Set notif popup position after size allocation

### LINKED ISSUE #139: Notification position incorrect after upgrade
## Description
Since upgrading Ubuntu Budgie from 21.10 to 22.04 (not sure which version of Budgie Desktop that involves), I have a minor problem with the position of notifications.
I have them set to "bottom right", but they now appear a few inches up the screen instead of in the corner as they used to. Still works fine, just looks odd.

## Budgie version
10.6.1 on Ubuntu Budgie 22.04

## Operating system
Ubuntu Budgie 22.04

Set notification position to "bottom right"
Receive a notification

## Actual result
Notification appears further up the screen than it should
image

## Expected result
Notification appears at the bottom right of the screen

## Additional information
I don't use any DPI scaling. I use 2540x1440 resolution. Fractional scaling is off.

serebit:
## Description

Instead of using GTK's estimation for the size our popup window will be, we attach to the size_allocate signal and show the popup before setting its position. Once size_allocate is called, the proper size of the notification is known, and we can immediately move it to the correct position before the screen is updated with the notification visible. This makes the behavior of the bottom-left and bottom-right notification positions match the behavior of the top-left and top-right positions.

Fixes #139.

### Submitter Checklist

### DIFF
--- a/src/daemon/notifications/dbus.vala
+++ b/src/daemon/notifications/dbus.vala
@@ -243,7 +243,6 @@
 				this.popups[id] = new Popup(this, notification);
 				this.configure_window(this.popups[id]);
 				this.latest_popup_id = id;
-				this.popups[id].show_all();
 				this.popups[id].begin_decay(notification.expire_timeout);
 
 				this.popups[id].ActionInvoked.connect((action_key) => {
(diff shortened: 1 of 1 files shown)


EbonJaeger (approved):
Looks good to me, thanks!

PR #146, opened 2022-06-21
fix broken C std for gvc subproject

eli-schwartz:
Historically, Meson didn't know how to set c_std for a subproject at all. Meson 0.63.0 adds support for this, so the default_options kwarg to subproject now begins to respect this.

budgie-desktop uses c11, but artificially sets c89 for the gvc submodule. However, this does not actually work... the build then errors out when trying to build `subprojects/gvc/test-audio-device-selection.p/test-audio-device-selection.c.o`

Remove the faulty argument.

Fixes regression in commit 3d6d71c167dd9742cd072e7e6a3453c23ebb2658.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -130,7 +130,6 @@ with_polkit = get_option('with-polkit')
 # Get gvc built before we do anything
 gvc = subproject('gvc',
     default_options: [
-        'c_std=c89',
         'static=false',
         'pkglibdir=@0@'.format(rpath_libdir),
         'package_name=' + meson.project_name(),


eli-schwartz:
> Meson 0.63.0 adds support for this

Note that the version I specified is not packaged in *almost any* distributions that ship Budgie... because it is a release candidate that was released 2 days ago in order to give people the chance to test regressions.

(It is packaged by Debian experimental.)

This issue was caught during a mass rebuild in a private run, as part of the regression testing preparation for the final release. The idea is that we don't want Budgie (or other projects) to actually start failing builds, hence fixing this issue before final release. :)

serebit:
Entirely reasonable. If Meson 0.63.0 adds support for the argument, does it build with c11 specified instead of c89?

eli-schwartz:
Per my commit message, all past versions of Meson build the subproject with c11 (and this is successful, of course) because they use the std from the main project (budgie).

You don't need to specify anything if you just want subprojects to use the same std as the main project. That used to be the only option, and with Meson 0.63.0 it's still the default option.

That being said, yes, it compiled correctly for me with this change and Meson 0.63.0 (and an additional change to not fatally error out when I have neither budgie nor gnome versions of the screensaver, as I'm not actually interested in building and installing additional useless things in order to test-compile a project I don't personally use).

More specifically, it compiled with the same GCC command line as Meson 0.62.2.

serebit:
If all this does is maintain (albeit unintended) behavior from prior meson versions and unbreak the build on future versions, then I'm fine with merging this, given that the Action has also passed. Thanks for the contribution.

fossfreedom:
@eli-schwartz much appreciation for this fix - this resolves a FTBFS in debian unstable.

PR #154, opened 2022-07-10
Fix PowerIndicator spam by only emitting warning once.

### LINKED ISSUE #151: PowerIndicator spamming journal log every 30 seconds
## Description
Running on virtualbox I was perusing the journal and noticed every 30 seconds our PowerIndicator was spamming the log
Screenshot from 2022-06-29 14-02-18

Logically makes sense in a virtualbox session where there isnt a "battery device" as such.  Not sure what happens on a desktop machine as opposed to a laptop where there similarly is likely not to be a battery to report on.

For a lengthy virtualbox session / non-battery based desktop the journal will grow unnecessarily fast.

Solution?  Maybe don't bother reporting / write once only and never again / any other suggestion?

## Budgie version
10.6.1

## Operating system
UB 22.10

Run budgie in a virtualbox session

## Actual result
See description

## Expected result
Suggest only write once or don't bother.

JoshStrobl:
Fixes #151

### DIFF
--- a/src/applets/status/PowerIndicator.vala
+++ b/src/applets/status/PowerIndicator.vala
@@ -13,6 +13,7 @@ public class BatteryIcon : Gtk.Box {
 	/** The battery associated with this icon */
 	public unowned Up.Device battery { protected set; public get; }
 	bool changing = false;
+	bool emitted_warning = false;
 
 	private Gtk.Image image;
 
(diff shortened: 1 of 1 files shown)


fossfreedom (approved):
Testing on virtualbox. Works as expected (i.e. watched journalctl for a while).

Cheers!

PR #157, opened 2022-07-11
Fix Budgie Menu view not resetting on open

EbonJaeger:
## Description

The previous `show()` function was only ever actually called when the widget is first shown. This adds a new abstract function for views to implement that gets called whenever the menu is opened, so any cleanup actions can be reliably performed.

### Submitter Checklist


JoshStrobl (approved):
Thanks for the fix!

PR #168, opened 2022-07-26
IBUS Rework

fossfreedom:
## Description
This maybe specific to Debian/Ubuntu - so it could be construed as a FYI as to what I will be adding as a patchset for these distros.  Thus please consider carefully - it does need thorough testing on your specific distro since as far as I can gather many distro's patch IBUS stuff in different ways. I don't recall seeing previous issues like this being reported by other distros hence the issues below may be specific to Debian/Ubuntu.

Over the years of budgie-desktop various ubuntu community members (and reproduced on Debian) have reported on occasions that the keyboard layout always defaulted to en_US even if it wasn't installed.  Logging in and logging out may resolved.  On other occasions various workaround like toggling the gsettings keyboard layouts key seem to resolve for a session.

Additionally various people have reported that ALT+Shift rarely worked as a switch layout key combination.

Finally - the IBUS popup for non-latin languages didn't appear in budgie.  We resolved this a while back and have carried the fix as a distro specific patch (the ibus-daemon process needed to be called with different parameters).

Screenshot from 2022-07-26 09-44-01

This PR reworks budgie's IBUS handling to resolve all of the above.  

As a FYI - I will be carrying this as a patchset for 22.10 where the first issue always seems to occur now - at least for me.  Various team members and community members are testing this for 22.04 and will likely push this as a stable release update for that series.  If demand warrants this I will also consider backporting to 20.04 where these issues have also been reported in the past.

### Submitter Checklist


src/wm/ibus.vala
```diff
@@ -73,7 +73,7 @@ namespace Budgie {
 		* Launch the daemon as a child process so that it dies when we die
 		*/
 		private void startup_ibus() {
-			string[] cmdline = {"ibus-daemon", "--xim", "--panel", "disable"};
+			string[] cmdline = {"ibus-daemon", "--xim", "--daemonize"};
```
JoshStrobl, line 76: Is daemonizing going to put the ugly ibus icon into the system tray permanently? There has been a back-and-forth over the years about enabling the panel option and it always comes down to a system tray item sticking out like a sore thumb.

src/wm/keyboard.vala
```diff
@@ -73,20 +76,49 @@ namespace Budgie {
...
+			/* Hook into GNOME defaults */
+			var schema = new Settings("org.gnome.desktop.wm.keybindings");
+			wm.get_display().add_keybinding("switch-input-source", schema, Meta.KeyBindingFlags.NONE, switch_input_source);
+			wm.get_display().add_keybinding("switch-input-source-backward", schema, Meta.KeyBindingFlags.NONE, switch_input_source_backward);
+		}
+
+		construct {
+			var schema = GLib.SettingsSchemaSource.get_default ().lookup ("org.gnome.desktop.input-sources", true);
+			if (schema == null)
+				return;
```
JoshStrobl, line 106: Please use curly braces.

src/wm/keyboard.vala
```diff
@@ -73,20 +76,49 @@ namespace Budgie {
 		/* Guard ourselves from any future potential derps */
 		private bool is_keyboard_held = false;
 
-		public KeyboardManager(Budgie.BudgieWM? wm) {
-			Object(wm: wm);
+		public static void init (Budgie.BudgieWM? wm) {
+			if (instance != null)
+				return;
```
JoshStrobl, line 81: Please use curly braces.

src/wm/keyboard.vala
```diff
@@ -310,12 +349,72 @@ namespace Budgie {
...
+		bool handle_modifiers_accelerator_activated (Meta.Display display) {
+			display.ungrab_keyboard (display.get_current_time ());
+
+			var sources = settings.get_value ("sources");
+			if (!sources.is_of_type (sources_variant_type))
+				return true;
+
+			var n_sources = (uint) sources.n_children ();
+			if (n_sources < 2)
+				return true;
```
JoshStrobl, line 374: Please use curly braces.

src/wm/keyboard.vala
```diff
@@ -310,12 +349,72 @@ namespace Budgie {
...
+		/**
+		* Respond correctly to ALT+SHIFT_L
+		*/
+		[CCode (instance_pos = -1)]
+		bool handle_modifiers_accelerator_activated (Meta.Display display) {
+			display.ungrab_keyboard (display.get_current_time ());
+
+			var sources = settings.get_value ("sources");
+			if (!sources.is_of_type (sources_variant_type))
+				return true;
```
JoshStrobl, line 370: Please use curly braces.

src/wm/keyboard.vala
```diff
@@ -310,12 +349,72 @@ namespace Budgie {
...
+			return true;
+		}
+
+		/**
+		* Called whenever  the keyboard layout needs to be set/reset
+		*/
+		[CCode (instance_pos = -1)]
+		void set_keyboard_layout (string key) {
+			if (!(key == "current" || key == "sources" || key == "xkb-options"))
+				return;
```
JoshStrobl, line 387: Please use curly braces.

src/wm/keyboard.vala
```diff
@@ -310,12 +349,72 @@ namespace Budgie {
...
+
+			string layout = DEFAULT_LAYOUT, variant = DEFAULT_VARIANT, options = "";
+
+			var sources = settings.get_value ("sources");
+			if (!sources.is_of_type (sources_variant_type))
+				return;
```
JoshStrobl, line 393: Please use curly braces.

src/wm/keyboard.vala
```diff
@@ -310,12 +349,72 @@ namespace Budgie {
...
+
+			unowned string? type = null, name = null;
+			if (sources.n_children () > current_source)
+				sources.get_child (current_source, "(&s&s)", out type, out name);
```
JoshStrobl, line 397: Please use curly braces.

JoshStrobl (commented):
Need to review the rest later but some initial items and question.

src/wm/ibus.vala
fossfreedom: interesting question.  On UB we override the dconf key so it doesn't display if the system tray is used. We don't carry the system tray as part of our default layout anyway - so a bit belt and braces.  For some layouts such as Japanese, if the user is NOT using a japanese keyboard with the extra keys to change the input mode, the only way to change the input mode (at least that I can find) is to re-enable the system tray icon (via ibus-setup) and change the input mode that way.

As far as I can gather on Debian/Ubuntu, the popup is a necessity for non-latin layout users which is a  little surprising statement I suppose since I haven't seen any previous comments on non debian/ubuntu forums for the lack-of.

Why the keyboard layout applet doesnt handle input modes is another question - and a different issue.

JoshStrobl: > We don't carry the system tray as part of our default layout anyway - so a bit belt and braces.

Sure but it *is* the default for Budgie itself, which is what should be tested against :)

> On UB we override the dconf key so it doesn't display if the system tray is used

Gotcha. Yea this sounds then like it is more something that needs to be decided on by anyone shipping Budgie Desktop (like UB does).

serebit: Can it not be launched with both options?

fossfreedom: the --panel disable hides it from the system tray.  But then there is no means to re-enable it.  Certain non-latin layouts such as japanese you need a means to change the input mode (if the keyboard you are using doesnt have the input mode key) which our current keyboard applet does not support.  How to extend the applet - no idea off the top of my head.

serebit: Could include an override in carbontray to hide the icon, with an associated applet option to unhide it.

JoshStrobl: @serebit That is a good idea.

JoshStrobl: Would be nice to see this get implemented as part of this PR.

JoshStrobl (changes requested):

usergreen:
> fossfreedom commented on Jul 26
>
>For some layouts such as Japanese, if the user is NOT using a japanese keyboard with the extra keys to change the input mode, the only way to change the input mode (at least that I can find) is to re-enable the system tray icon (via ibus-setup) and change the input mode that way.

I think this problem is only specific to Ubuntu Budgie. Other distros including Debian and Fedora can change the input mode by pressing Shift + Space and/or Ctrl + Space keys besides the system tray icon.

I tested Ubuntu Budgie 22.10. I followed the Deian way to introduce IBus-anthy, but the system tray icon did not appear at all. So it seems that this might be a specific issue only for Ubuntu Budgie.

Best regards.

fossfreedom:
@usergreen you will need to add the system tray applet to the panel. You then need to enable the tray icon via ibus-settings. Only then will you see the ibus icon.

usergreen:
@fossfreedom As I mentioned above, there is no need to add the system tray applet to the panel in order to enable Japanese input. Of course, ibus tray applet is an available option to add afterwards.

In Ubuntu Budgie, you can switch input method from English to Japanese by Ctrl + j key (or 半／全 key in Japanese keyboard).

What I did to enable Japanese input was only to do the following:
`apt install ibus-anthy`
 as root user and
`im-config`
 as ubuntu-budgie user and choose IBus from the menu.
Then added  [System Tools] > [IBus Preferences] > [Imput Method] > [Add] > [日本語] > [Anthy]

The above can be reproduced.

fossfreedom:
Indeed. You shouldnt (in theory)  need ibus preferences at all.

Just use the appropriate keyboard layout in budgie-control-center

For example, for Chinese install ibus and lib-ibus-pinyin  logout and login and then choose the intelligent pinyin layout in BCC - keyboard. Logout and login again.

src/wm/ibus.vala
fossfreedom: Override added and as discusses a BDS option to toggle assuming ibus-daemon is installed and running.

JoshStrobl (approved):
Went ahead and did some code cleanup. LGTM. Thank you for adding the ibus icon option!

JoshStrobl (changes requested):
Correction: Cycling seems bork. It defaults to en + Finnish even though I've only specified Finnish in BCC (in ibus it seems to have en + Finnish), but I can't switch between the two. I can only switch to en US but not back. When I add Japanese (Anthy) in BCC, I can switch and it shows up in the tray, and I can switch to English, but I can't switch to Finnish. The Japanese Anthy layout doesn't appear in the popup / tooltip for IBUS either.

usergreen:
In that case, I suggest to use "uim" (Universal Input Method) instead of IBus. uim + Anthy works fine for me and has no problem. You can test my re-spin ISO image file from https://sourceforge.net/projects/mx-respin-iso/files/budgie/

Best regards.

fossfreedom:
Rather than just pointing to an iso please do help with this. Raise a discussion thread and describe what is uim, how you have implemented it with any hints and tips on setting things such as anthy et al, how you suggest the keyboard applet should flip between layouts I.e. what files/dconf key needs tweaking.  Thanks in advance.

JoshStrobl:
@fossfreedom should this be closed in favor of the other keyboard layout work you're doing?

PR #179, opened 2022-08-23
Change label properties for notification bodies

### LINKED ISSUE #171: Notification bodies are inconsistent
## Description

Notifications from some programs aren't ellipsized (KDE Connect, web browsers) and wrapped, while others are (`notify-send`). Very long messages could take up the entire height of the screen, and more.

I think this also affects the notifications displayed in Raven.

## Budgie version

budgie-desktop 10.6.2 (git-481817675bc27826a7cf9f39e209a27853fda75b)

## Operating system

Solus 4.3 Fortitude

1. Go to https://www.bennish.net/web-notifications.html in a web browser
2. Click the Authorize button to allow notifications
3. Click Show

For contrast,
1. In your terminal, run: `notify-send -a Test aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa`

## Actual result

image

## Expected result
image

## Additional information
I'm not sure what the Right Thing:tm: to do here even is. Do we want to limit to one line? Two lines? What's the best way to do multiple lines if so?

EbonJaeger:
## Description

This makes notification popup bodies more consistent, and prevents them from taking up unnecessarily large portions of the screen.

### Property Changes
The max chars were reduced slightly to ensure that the close button doesn't get shoved off the edge of the notification popup. Width chars were added so that the body text always has the same minimum and maximum size per line, so the entire popup body is filled by the label. Lines is set to 2, so only two lines of text is displayed. Note that this does not mean two lines in the popup; it's two lines of the actual notification text. This allows long lines to wrap (and wrapped lines don't count towards lines, because reasons) without popups taking up large amounts of screen real-estate. Xalign is set to 0 to ensure that the text always starts at the left (because halign and justify isn't enough).

### Screenshot

The top notification has a single long line that wraps, while the other consists of multiple lines.

image

It would be great for people to try this branch out for a day (or several) before merging, so any concerns and tweaks can be addressed and made using actual day-to-day notifications.

Fixes #171

### Submitter Checklist

### DIFF
--- a/src/daemon/notifications/popup.vala
+++ b/src/daemon/notifications/popup.vala
@@ -265,9 +265,11 @@ namespace Budgie.Notifications {
 				use_markup = true,
 				wrap = true,
 				wrap_mode = Pango.WrapMode.WORD_CHAR,
-				max_width_chars = 35,
-				halign = Gtk.Align.START,
+				width_chars = 33,
+				max_width_chars = 33,
+				lines = 2,
 				valign = Gtk.Align.START,
+				xalign = 0,
 				hexpand = true,
 				vexpand = true
 			};


JoshStrobl:
Been dogfooding for a while and haven't noticed any regressions, so merging :+1:

PR #185, opened 2022-08-30
Add missing gio-unix-2.0 dependencies

FedericoSchonborn:
## Description

This commit adds all missing `gio-unix-2.0` dependencies on all targets that require it. This provides the `-I${PREFIX}/include/gio-unix-2.0` compiler flag, which fixes the following error:

image

This issue was already reported to other projects by the NixOS community: https://github.com/NixOS/nixpkgs/issues/36468.

This patch can be applied to the tarball released with version 10.6.3: https://github.com/FedericoSchonborn/budgie-overlay/blob/f7b6e707d183fc3f8ee31efb923249252cb9910f/packages/budgie-desktop/0001-Add-missing-gio-unix-2.0-dependencies.patch (it omits the `appindexer` directory, which isn't included in release builds).

### Submitter Checklist

### DIFF
--- a/src/abomination/meson.build
+++ b/src/abomination/meson.build
@@ -8,6 +8,7 @@ libabomination_sources = [
 ]
 
 libabomination_deps = [
+    dep_giounix,
     dep_gtk3,
     dep_gdkx11,
     dep_wnck,
--- a/src/appindexer/meson.build
+++ b/src/appindexer/meson.build
@@ -11,6 +11,7 @@ libappindexer = shared_library(
     appindexer_sources,
     dependencies: [
         dep_gee,
+        dep_giounix,
         dep_gtk3,
     ],
     vala_args: [
--- a/src/applets/budgie-menu/meson.build
+++ b/src/applets/budgie-menu/meson.build
@@ -43,6 +43,7 @@ applet_budgiemenu_sources = [
 applet_budgiemenu_deps = [
     libplugin_vapi,
     dep_gee,
+    dep_giounix,
     dep_gtk3,
     dep_peas,
     dep_accountsservice,
--- a/src/applets/clock/meson.build
+++ b/src/applets/clock/meson.build
@@ -39,6 +39,7 @@ shared_library(
     applet_clock_sources,
     dependencies: [
         libplugin_vapi,
+        dep_giounix,
         dep_gtk3,
         dep_peas,
         link_libplugin,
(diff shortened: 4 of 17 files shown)


JoshStrobl:
Did you validate that this is needed for every single one of the meson.build files you added it to?

FedericoSchonborn:
> Did you validate that this is needed for every single one of the meson.build files you added it to?

I added the dependency target by target as builds were failing.

> And why is this seemingly a nixOS specific issue? Other operating systems sensibly pull up the glib devel headers when you pull in libgtk-3 devel headers.

No idea, it seems to be a common problem with Nix packages that require `gio-unix-2.0` (hence the linked PR), but I haven't found any explanation on why it happens.

eli-schwartz:
```
$ pkg-config --cflags --libs gtk+-3.0
-I/usr/include/gtk-3.0 -I/usr/include/pango-1.0 -I/usr/include/glib-2.0 -I/usr/lib/glib-2.0/include -I/usr/include/harfbuzz -I/usr/include/freetype2 -I/usr/include/libpng16 -I/usr/include/libmount -I/usr/include/blkid -I/usr/include/fribidi -I/usr/include/cairo -I/usr/include/lzo -I/usr/include/pixman-1 -I/usr/include/gdk-pixbuf-2.0 -I/usr/include/gio-unix-2.0 -I/usr/include/cloudproviders -I/usr/include/atk-1.0 -I/usr/include/at-spi2-atk/2.0 -I/usr/include/dbus-1.0 -I/usr/lib/dbus-1.0/include -I/usr/include/at-spi-2.0 -pthread -lgtk-3 -lgdk-3 -lz -lpangocairo-1.0 -lpango-1.0 -lharfbuzz -latk-1.0 -lcairo-gobject -lcairo -lgdk_pixbuf-2.0 -lgio-2.0 -lgobject-2.0 -lglib-2.0 
```

It's right there in the public cflags for gtk+-3.0, because it is listed in `Requires.private:` for `gtk+-3.0.pc` -- private dependencies provide their cflags, but not their libs (see various discussion e.g. https://bugs.freedesktop.org/show_bug.cgi?id=105572).

- It sounds like NixOS doesn't have Requires.private "correctly" specified, per the upstream generated gtk pkg-config files.
- Technically, gtk itself doesn't expose gio-unix-2.0 anyway, it only needs that for `Requires.internal`... which the freedesktop pkg-config maintainers repeatedly refused to implement.


The underlying issue is that budgie itself uses `gio-unix-2.0` ***directly***, but doesn't list it ***directly*** as a dependency and instead assumes that it will be transitively included by `gtk+-3.0`.

You do pass `--pkg gio-unix-2.0` in the vala_args in such cases.

Issue #183, opened 2022-08-30
Workspace Switcher: The icon is flickering very fast when the cursor on it


Solarunit:
## Description

The icon is flickering very fast when the cursor on it (also in the popover)

![](https://i.imgur.com/LXNg5wl.png)

## Budgie version

budgie-desktop 10.6.3

## Operating system

Solus

## Steps to reproduce the issue

Hover the cursor over the icon.

Issue #182, opened 2022-08-30
Workspace Switcher: Drag&drop doesn't work


Solarunit:
## Description

Drag&drop doesn't work in the Workspace Switcher applet (when trying to drag icon from workspace-1 to  workspace-2).
It also doesn't work when dragging fom the popover

![](https://i.imgur.com/LXNg5wl.png)

## Budgie version

budgie-desktop 10.6.3

## Operating system

Solus

## Steps to reproduce the issue

Drag&drop icons from one Workspace to another.

serebit:
I didn't know that drag and drop in the workspaces applet was a thing, and I can't drag and drop in v10.6.3 while @fossfreedom is able to drag and drop in v10.6.1, so marking as verified.

Issue #180, opened 2022-08-23
Workspace Switcher High Memory Usage in Budgie 10.6.3


GZGavinZhao:
## Description

`budgie-panel` consumes a lot of memory when Workspace Switcher is added to the panel.

## Budgie version

10.6.3. I rolled back to 10.6.2 and the problem disappeared, so it seems like 10.6.3 related.

## Operating system

Solus 4.3

## Steps to reproduce the issue

1. Open any system monitor, find the single `budgie-panel` process.
2. Open Budgie Desktop Settings. In Panels, click "add applet" and add a Workspace Switcher to any location in any panel.
3. Watch system monitor and check

## Actual result
The memory usage should not increase (a lot), as in Budgie 10.6.2.

## Expected result
The memory usage continues to increase and doesn't show signs of stopping. An extreme case brought this usage to 9GB.

Also, when I hover my mouse on any of the icons, it flashes repeatedly instead of just lighting up once.

## Additional information
#150 might be related since the reported user also has Workspace Switcher in his applet list and this issue went unresolved. However, he's on 10.6.1, so it also might not be related :P

serebit:
#150 was due to a third-party applet, as detailed in the comments prior to the closure of the issue. I'll check this out and see if I can find anything.

nathanpainchaud:
I didn't notice the high memory usage, but I **did** notice the icons flashing when hovering over them. Also, I'm used to dragging the icons to move programs between workspaces, and that's not working anymore either. I rolled back to 10.6.2 and everything is working as intended.

JoshStrobl:
@nathanpainchaud Those are separate issues and should be filed separately. Deleting comment to keep this on topic.

nathanpainchaud:
@JoshStrobl It's an issue with the workspace switcher mentionned by the original poster himself, so I thought it was relevant to mention, in case it could have helped @serebit figure out what's going on. But I'll let you be the judge of that...

PR #187, opened 2022-08-30
Resolve numerous issues with the Workspace Applet

### LINKED ISSUE #184: Workspace Switcher: right click menu doesn't work
## Description

Right click menu doesn't work, it could appear for a a fraction of a second then vanishes.

## Budgie version

budgie-desktop 10.6.3

## Operating system

Solus

Right click in the Workspace Switcher applet

serebit:
## Description
Replaces a `return true` with a `return false` in a critical Timeout callback so the function isn't repeated every 100ms.

Fixes #184, #183, #182, and #180.

### Submitter Checklist


JoshStrobl (approved):
LGTM, thanks!

PR #192, opened 2022-09-04
Run Dialog refactor and improvements

### LINKED ISSUE #89: Sort entries in Run Dialog depending on the search text
### Current Budgie version
10.6

### Use-cases
When using the run dialog and searching for a term, all non-matching applications are hidden. However, the results are not sorted "smartly" but are sorted alphabetically. This can be annoying since unnecessary amount of typing is required in order to pin down the application I search for.   
In the following image, I am searching for "firefox" and "fire" does not suffice. In my case, I need to add another "f" or use the key arrows before I can press enter.
screenshot_22_03_28_19_33_39

### Proposal
The search in the  budgie menu already implements "smart" sorting. From what I can make out, the relevant code for that is in [`do_sort_list`](https://github.com/BuddiesOfBudgie/budgie-desktop/blob/07007bf22d20ae1c336013db1baf11686ede9a84/src/applets/budgie-menu/BudgieMenuWindow.vala#L584) and [`get_score`](https://github.com/BuddiesOfBudgie/budgie-desktop/blob/07007bf22d20ae1c336013db1baf11686ede9a84/src/applets/budgie-menu/BudgieMenuButtons.vala#L113).  
This code should be easily portable to the run dialog. For example, [`set_sort_func`](https://valadoc.org/gtk+-3.0/Gtk.ListBox.set_sort_func.html) can be used similarly on [`app_box`](https://github.com/BuddiesOfBudgie/budgie-desktop/blob/07007bf22d20ae1c336013db1baf11686ede9a84/src/rundialog/RunDialog.vala#L128) to set a similar function to `do_sort_list` used in the budgie-menu code. Furthermore, the [`get_score`](https://github.com/BuddiesOfBudgie/budgie-desktop/blob/07007bf22d20ae1c336013db1baf11686ede9a84/src/applets/budgie-menu/BudgieMenuButtons.vala#L113) function can also be reused. For example, by deriving a new class from [`Gtk.ListBoxRow`](https://valadoc.org/gtk+-3.0/Gtk.ListBoxRow.html), adding `get_score` to it and using the derived class in place of `Gtk.ListBoxRow` in [`RunDialog.vala`](https://github.com/BuddiesOfBudgie/budgie-desktop/blob/07007bf22d20ae1c336013db1baf11686ede9a84/src/rundialog/RunDialog.vala).

EbonJaeger:
## Description

This implements using the new application indexer used by the new Budgie Menu to find and sort applications. When searching, applications are filtered and sorted just like the Budgie menu, taking fuzzy finding and relevancy into account.

To make sure that the Run Dialog closes as expected in all cases, the context signals for launching an application were hooked up and re-emitted. This allows the Run Dialog to use the same function for launching applications as the Budgie Menu, which enables setting GPU profiles if necessary for multi-GPU systems.

There are a couple of visual changes as well:

1. The width is now computed from the work area of the screen that the Run Dialog is opened on. This should make it look good on different sized monitors.

2. The label for the name and description of applications in the launcher buttons has been split into two labels. This allows more fine-grained styling than was previously possible.

Closes #89 

### Submitter Checklist


src/rundialog/LauncherButton.vala
```diff
@@ -0,0 +1,60 @@
...
+			margin_start = 8
+		};
+
+		var right_box = new Gtk.Box(Gtk.Orientation.VERTICAL, 6);
+
+		var app_name = Markup.escape_text(application.name);
+		var sdesc = application.description;
+		if (sdesc == null) {
+			sdesc = "";
+		}
```
JoshStrobl, line 21: What about: 

```
var sdesc = application.description ?? "";
```

JoshStrobl (changes requested):
Width could do with reduction as well. It is veeeeery wide. Otherwise LGTM!

image

src/rundialog/LauncherButton.vala
EbonJaeger: What even is that syntax lol. TIL

JoshStrobl: It's a nullish coalescence :) It will use any truthy value on the left and fallback to the value on the right if the value is falsey.

JoshStrobl (approved):
LGTM! :gem:

PR #195, opened 2022-09-18
Add extra check in XDG dir detection

serebit:
## Description

For some reason, on Arch derivatives, XDG user dirs that aren't present or that resolve as symlinks are detected - but they resolve to the user's home dir instead. This compares the result of get_user_special_dir to the user's home dir path, and skips the XDG dir if they're equal.

### Submitter Checklist


JoshStrobl (approved):
LGTM!

PR #200, opened 2022-09-25
Remove numeric space in formatted clock applet output

serebit:
This was introduced in the glib2 pull request [!2698](https://gitlab.gnome.org/GNOME/glib/-/merge_requests/2698) and is not desired behavior for us.

## Description
Previous behavior restored by this PR:
image
Behavior introduced by the glib2 commit:
image

### Submitter Checklist


JoshStrobl (approved):
LGTM, thanks!

Issue #176, opened 2022-08-15
Unreliable right-click behavior on window title bars, context menu often does not appear


ghost:
## Description
Right-clicking the title bar is unreliable.  It sometimes takes numerous right-click attempts for the context menu to appear.  This bug has been present for quite some time now.  

Issue on the old tracker: https://github.com/solus-project/budgie-desktop/issues/1116

Possibly related to #204 

## Budgie version
10.6.2

## Operating system
EndeavourOS

## Steps to reproduce the issue
Right-click a window title bar.  

## Actual result
The context menu often doesn't appear, and may take several right-clicks.

## Expected result
Context menu should appear on every right-click

serebit:
Yeah, I'm also seeing this on EndeavourOS. Marking as verified.

JoshStrobl:
Yea, doesn't like it when the window isn't in focus, but if the window is not in focus and you right-click on the context menu, it's fine.

dlangerm:
Same here on Ubuntu Budgie. Thought it was just my multi-monitor configuration causing issues.

0x199x:
> Yea, doesn't like it when the window isn't in focus, but if the window is not in focus and you right-click on the context menu, it's fine.

The window focusing doesn't change anything for me.  The menus won't come up whether the window is in-focus or out-of-focus.  

I also have the issue of not being able to right-click on my desktop either though, so there could be more going on on my end.

tomocafe:
Just adding a data point, I also lost this ability, not sure when. Though I can't _ever_ get the context menu to appear with the window focused or not.

```
System:
  Host: *** Kernel: 5.15.61-217.current arch: x86_64 bits: 64
    Desktop: Budgie v: 10.6.4 Distro: Solus 4.3 fortitude
CPU:
  Info: 8-core model: AMD Ryzen 7 1700 bits: 64 type: MT MCP
Graphics:
  Device-1: NVIDIA GP106 [GeForce GTX 1060 6GB] driver: nvidia v: 515.65.01
  Display: x11 server: X.Org v: 1.20.14 with: Xwayland v: 22.1.3 driver: X:
    loaded: nvidia gpu: nvidia resolution: 3840x2160~60Hz
  OpenGL: renderer: NVIDIA GeForce GTX 1060 6GB/PCIe/SSE2 v: 4.6.0 NVIDIA
    515.65.01
```

fossfreedom:
[...]

PR #207, opened 2022-10-25
Use idle and menu popup to display a Wnck ActionMenu

fossfreedom:
## Description
Draft - makes the window actionmenu display for most windows.

https://github.com/BuddiesOfBudgie/budgie-desktop/issues/176

Note - it uses the now deprecated Gtk.Menu.popup - I haven't figured out how to use one of the recommended alternatives.

ActionMenu doesn't appear on all windows - for example firefox esr or visual code (snap?)

We should also perhaps investigate removing the move workspace left/right menu items - should be able to-do that via the Gtk.Container remove methods at a guess. Remember the Wnck.ActionMenu is just a Gtk.Menu object

Anyway - marking this as draft.  **Feel free to improve** and eventually mark as ready.

### Submitter Checklist

### DIFF
--- a/src/daemon/menus.vala
+++ b/src/daemon/menus.vala
@@ -119,7 +119,7 @@ namespace Budgie {
 				if (desktop_menu.get_visible()) {
 					desktop_menu.hide();
 				} else {
-					desktop_menu.popup_at_pointer(null);
+					popup_menu(desktop_menu, button, timestamp);
 				}
 				return false;
 			});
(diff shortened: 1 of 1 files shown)


serebit (approved):
Looks fine and works fine on my end, and though it doesn't entirely fix the longstanding issues with the window menu, it does fix the desktop right-click menu. If there are no objections, I'll merge this tomorrow.

JoshStrobl:
> it doesn't entirely fix the longstanding issues with the window menu, it does fix the desktop right-click menu.

IMO this was the more critical of the issues.

Issue #208, opened 2022-10-27
Dynamic panel is ignored with fractional scaling enabled


fossfreedom:
## Description
With fractional scaling set at 125%, when a window is maximised, the dynamic panel colouring is ignored - i.e. it remains transparent.

If I maximize on my second screen, the first screen with the panel is then coloured correctly.

So the issue here is that scaling with recognising when a window is maximised is improperly calculated for the primary screen for setting the dynamic CSS property.

I'll assign myself to look at this at some point unless someone in the meantime grabs the issue.

## Budgie version
10.6.4 

## Operating system
Ubuntu Budgie 22.10

## Steps to reproduce the issue
set fractional scaling to 125% using budgie-control-center, logout and login and maximise any application - remember to ensure the panel is set to dynamic.

## Actual result
remains transparent

## Expected result
should be coloured.

PR #211, opened 2022-11-02
Ensure calculation for windows on the primary monitor include monitor…

fossfreedom:
… scale factors - closes: #208

## Description
The algorithm that determines whether a window intercepts a panel (such as raven or a maximised window) ignores the scale factor of the panel's primary monitor.

Thus it gets confused when two monitors have differing scaling.
This PR ensures that the primary monitor scale factor is used when calculating intercept which is the decision point used to switch from opaque to transparent when the panel is set to dynamic.

### Submitter Checklist


JoshStrobl (approved):
Noticed no regressions in testing. LGTM!

Issue #117, opened 2022-05-01
What are our supported Notification options


fossfreedom:
## Description
In budgie control center we have this dialog for each application

![image](https://user-images.githubusercontent.com/996240/166142768-df70f9ce-1fc6-42f9-823e-12a1fc443bbf.png)

End user is reporting that

## Budgie version
BCC 1.0
Budgie Desktop v10.6.1

## Operating system
Ubuntu Budgie 22.04

## Steps to reproduce the issue
Enable notification sound for an application e.g. network applet
Wait until that app produces a notification - e.g. disconnect / reconnect a network

## Actual result
No sound for a notification

## Expected result
That's the question ... if we don't support sounds then the BCC dialog needs to be adjusted.  What needs to be removed?
If we do - then we need to tweak the notifications stuff in budgie

EbonJaeger:
At the moment, only Notifications and Notification Popups are checked. I think supporting the Show Message Content option would be worthwhile. It shouldn't be too difficult to support sounds as well with the new notifications stuff, either.

fossfreedom:
Ah. Excellent.  Cheers.

PR #222, opened 2022-11-23
More notification enhancements

EbonJaeger:
## Description
This PR contains several fixes and enhancements to the notification system. Here is a list of the user-facing changes:

- #### Fix large fonts pushing the close button out of the popup
  _This sets the notification body to use the size of the widget for wrapping instead of a character length. Notification titles also were given a margin-right to prevent overlapping the close button._

- #### Add an animation to notification popup opening and closing
  _Notification popups will now fade in and out upon showing and closing._

- #### Prevent notification popups from stealing input focus

- #### Only show three notification popups on the screen at a time
  _If there are more than three notifications in a short amount of time, the overflow ones will be sent directly to Raven._

- #### Sort notifications in Raven from old to new
- #### Implement support for notification sounds
  _Notification popups will now play a sound to alert the user. Playing a sound for a notification can be enabled or disabled in Budgie Control Center for each individual application._

- #### Show a notification when showing alerts are unpaused
  _When a window leaves fullscreen mode, Budgie will now show a notification popup to alert the user that there are notifications that they didn't see. This notification is only shown if there are actually notifications._

- #### Don't show notification content in popups if not configured to do so

### Packagers

This PR adds a dependency on `libcanberra` and `libcanberra-gtk3`. These may or may not already be installed as dependencies of other packages in the stack.

A new gschema override is also shipped so that notification content is shown by default. This is done because the upstream default for this setting is to not show, because GNOME and other desktops using this setting show the notification content on mouse hover, whereas we show it at all times.

### Submitter Checklist


serebit (changes requested):
To address each change:

- Large fonts no longer push the close button out of the popup on my laptop.
- The animation works fine on my system.
- Wasn't able to test this.
- This works, though I think it would be nice to have a setting to control how many notifications can be shown at once.
- Sorting in Raven mostly works, though I did find an edge case where it doesn't - if a bunch of notifications are sent at the same time, the notifications that were shown on screen are sorted after the ones that were buffered, despite being first.
- Notification sounds do work! Unfortunately, though, most applications that can cause notification sounds on my system don't show up in Budgie Control Center, so there's no way to disable the sounds for those applications. That may need to be addressed in B-C-C.
- Haven't tested this yet.

I also think that it may be wise to show "and X additional notifications" when more notifications than can be shown are submitted to the daemon, rather than silently sending them over to Raven. It's a bit confusing when some notifications show up and some don't, and there's no visual or auditory indication that notifications were sent that didn't appear.

EbonJaeger:
> Notification sounds do work! Unfortunately, though, most applications that can cause notification sounds on my system don't show up in Budgie Control Center, so there's no way to disable the sounds for those applications. That may need to be addressed in B-C-C.

Yeah, definitely. I think we (or at least Josh and I) agreed that we should maybe move the notification settings in BCC from the current Notifications panel into the Applications panel, which *should* have all the applications listed.

serebit:
Said this in the Matrix chat, but I'll say it here too—I'd also recommend implementing the Show Message Content setting in #117 so we can close that issue.

JoshStrobl:
I think the experience of getting notification or double sounds is going to be quite frustrating without the work in BCC that enables turning it off for more applications. Any thoughts on the default being off and only turning it on for applications where it's explicitly enabled through BCC, until the groundwork in BCC is done?

EbonJaeger:
Yeah, that sounds good.

JoshStrobl (approved):
LGTM :gem:

serebit (approved):

PR #228, opened 2022-12-06
Fix more warnings

serebit:
## Description

This patch resolves two warnings:

- Warning output by Meson due to a missing `check` argument in the run_command call
- Missing `void` argument for some C functions (warning output by Clang 15)

### Submitter Checklist


JoshStrobl (approved):
LGTM

PR #230, opened 2022-12-12
Update workspace item icons when window icon changes

### LINKED ISSUE #224: Workspace Switcher: In some file managers the icon doesn't match the current directory
## Description

Hello,
When using file managers like [Nemo](https://github.com/linuxmint/nemo) and [Thunar](https://docs.xfce.org/xfce/thunar/start), the icon in the workspace switcher applet shows a representation of the current directory the user is in, instead of the generic application icon:
Selection_003

(Here with the Papirus Dark Icon Theme)

But this icon does not change on the fly when switching folders, so it's possible to have cases like these, where you get conflicting information:

Untitled4

The Video-Icon is somehow not updated to the Templates icon, although I changed locations. This icon seems to be only redrawn, if you minimize the application, or switch applications. 

Is it possible to update the workspace switcher icon always to the current folder directory? If it's not technically feasible, it might be preferable to always force the generic application icon on Nemo / Thunar, instead of the symbolic icon.

## Budgie version

10.6.4

## Operating system

Arch Linux x86_64

Install Nemo / Thunar, activate the Workspace Switcher applet and change folders in the users home directory

## Actual result

Workspace Switcher shows mismatching symbolic icons instead correct ones / the generic application icon

## Expected result

Icons in the Workspace Switcher should always match current directory, or just show the generic Application Icon

## Additional information

Not every file manager has this problem, Nautilus works just fine.
There may be more file managers with this problem.

serebit:
## Description

Fixes #224.

### Submitter Checklist

### DIFF
--- a/src/applets/workspaces/WindowIcon.vala
+++ b/src/applets/workspaces/WindowIcon.vala
@@ -29,6 +29,11 @@ namespace Workspaces {
 				this.set_tooltip_text(window.get_name());
 			});
 
+			window.icon_changed.connect(() => {
+				icon.set_from_pixbuf(window.get_mini_icon());
+				icon.queue_draw();
+			});
+
 			Gtk.drag_source_set(
 				this,
 				Gdk.ModifierType.BUTTON1_MASK,


JoshStrobl (approved):
LGTM, thanks!

PR #231, opened 2022-12-18
update copyright notice

mrhappyma:
This was randomly irritating me so here I am. Unless there's some reason why it should be left at 2021, which there very well could be.


fossfreedom (approved):
Thx. There is going to be a mass update shortly for 2023 anyway.

PR #235, opened 2022-12-24
Add Meson options to allow building on OpenBSD

bilaliscarioth:
Adds two Meson options (`with-bluetooth` and `with-libuuid-time-safe`) to allow building on OpenBSD. By default, OpenBSD is missing Bluetooth for security reasons, and `generate_time_safe` is unavailable.


src/panel/uuid.vala
```diff
@@ -28,7 +28,10 @@ namespace LibUUID {
 		} else if ((flags & LibUUID.UUIDFlags.TIME_TYPE) != 0) {
 			LibUUID.generate_time(time);
 		} else if ((flags & LibUUID.UUIDFlags.TIME_SAFE_TYPE) != 0) {
-			LibUUID.generate_time_safe(time);
+#if USE_OPENBSD
```
serebit, line 31: Should move the macro one line up, to be above the "else if" rather than below it. As it stands, if built for OpenBSD and the TIME_SAFE_TYPE flag is passed, nothing will be executed.

src/applets/status/meson.build
```diff
@@ -38,6 +39,19 @@ applet_status_deps = [
     gvc.get_variable('libgvc_dep'),
     meson.get_compiler('c').find_library('m', required: false),
 ]
+else
+applet_status_deps = [
```
serebit, line 43: It would probably be better to assign the array without the gnome-bluetooth dep, then add it to the array if the host machine isn't openbsd, rather than having two separate assignments.

src/applets/status/BluetoothIndicator.vala
```diff
@@ -21,7 +21,10 @@ public class BluetoothIndicator : Gtk.Bin {
 	public Gtk.Image? image = null;
 
 	public Gtk.EventBox? ebox = null;
+#if USE_OPENBSD
```
serebit, line 24: With regards to the Bluetooth stuff, I think it would be preferable to add a more general Meson option: `with-bluetooth`, defaulting to `true`. Then any operating system that doesn't ship gnome-bluetooth can set that option.

serebit (changes requested):
Ultimately, I think these changes should be modified to be two Meson options: `with-bluetooth`, and `with-libuuid-time-safe` (both defaulting to true). Having a code path specific to OpenBSD just doesn't seem right when I'm sure other operating systems could make use of these two options.

bilaliscarioth:
what do you thin with the new commit ?

src/panel/uuid.vala
```diff
@@ -27,12 +27,13 @@ namespace LibUUID {
...
-#if USE_OPENBSD
-#else
+		}
+#if with_libuuid_time_safe
+		else if ((flags & LibUUID.UUIDFlags.TIME_SAFE_TYPE) != 0) {
 				LibUUID.generate_time_safe(time);
+		}
 #endif
-		} else {
+		else {
```
serebit, line 36: Should stay `} else {`.

serebit (changes requested):
Looks good - just one style regression that I noticed.

serebit:
Also, latest commit fails to build. Looks like a typo in one of the `meson.build` files.

bilaliscarioth:
"Also, latest commit fails to build. Looks like a typo in one of the meson.build files."

Sorry i missed a char on 'applet_status_deps'  so it give an error T-T.

serebit (approved):
Looks good to me, though ultimately Josh has the final say. Thanks for the contribution!

src/panel/uuid.vala
```diff
@@ -27,8 +27,10 @@ namespace LibUUID {
 			LibUUID.generate_random(time);
 		} else if ((flags & LibUUID.UUIDFlags.TIME_TYPE) != 0) {
 			LibUUID.generate_time(time);
+#if with_libuuid_time_safe
 		} else if ((flags & LibUUID.UUIDFlags.TIME_SAFE_TYPE) != 0) {
-			LibUUID.generate_time_safe(time);
+				LibUUID.generate_time_safe(time);
```
JoshStrobl, line 32: Fix indentation change otherwise LGTM.

```suggestion
			LibUUID.generate_time_safe(time);
```

JoshStrobl (changes requested):

JoshStrobl (approved):
LGTM!

PR #236, opened 2022-12-25
Implement CI for armv7h, aarch64, and riscv64

serebit:
## Description

Adds stages to the CI workflow for armv7, aarch64 (armv8), and riscv64. All build on Ubuntu 22.04. The repository configuration will need to be updated to require the newly updated checks.

### Submitter Checklist

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -38,7 +38,8 @@ jobs:
       with:
         arch: ${{ matrix.arch }}
         distro: ${{ matrix.distro }}
-        run: sudo apt update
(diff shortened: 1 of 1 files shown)


JoshStrobl:
Rejecting. I don't want CI / CD to balloon in times for hardware we don't directly support, is validated by third-parties and their infra already (e.g. ARMv7h for Pi support), or for entire architectures where sufficient hardware is not current available (e.g. SiFive RISC-V boards). In the future this should be discussed before that work is put into it.

serebit:
Reopening as per discussion on Matrix. Will evaluate the benefit of dependency caching.

serebit:
I'm going to go ahead and say that this isn't worth it for budgie-desktop. When caching actually works, it slashes the time to build in half, but it doesn't always work. Unfortunate that this experiment didn't work out.

PR #243, opened 2022-12-29
Ensure screenshot man-page reflects revised usr/bin binary name

fossfreedom:
## Description
Our screenshot binary name was revised during its PR passage.  

Its man-page needs to be similarly revised to reflect the new reverse DNS naming convention

Tested on ubuntu Jammy and kinetic

### Submitter Checklist


JoshStrobl (approved):
LGTM, thanks!

PR #244, opened 2023-01-01
Implement power widget

EbonJaeger:
## Description
Adds a new application that runs in the background that, when shown, contains a grid of buttons to allow the user to lock the screen, log out, suspend, shutdown, etc.

The Power Dialog is shown or hidden via DBus. The Budgie Menu and User Indicator applets now toggle this new dialog instead of implementing their own separate UI/UX, reducing duplicate code and unifying the logic all in one place.

A keybind has been added to open the Power Dialog (default: `<Super>` + `<Shift>` + `P`). A separate PR to Budgie Control Center will be done to have this shown in the keybinds section.

### CSS Classes
There are several new CSS classes to make theming the dialog easier. See the docs on PowerWindow in `window.vala` to see the structure and classes.

### Submitter Checklist


src/theme/common/_dialogs.scss
```diff
@@ -44,6 +49,28 @@
     }
 }
 
+// Power Dialog
+@mixin budgie_power_dialog($button) {
```
JoshStrobl: $button isn't used.

```suggestion
@mixin budgie_power_dialog() {
```

JoshStrobl (approved):
LGTM! Fantastic work :gem:

PR #246, opened 2023-01-02
Fix orientation with dual batteries

### LINKED ISSUE #245: Status Indicator applet displays multiple batteries oddly on a vertical panel
## Description
Battery icons / percentage labels display odd on systems with dual batteries (i.e. a laptop with both an internal and external battery)

## Budgie version
Version 10.6.4

## Operating system
Ubuntu Budgie 22.10

On a system with two batteries, make sure "Show battery percentage" is enabled.  Switch the panel to be left or right. Batteries are displayed horizontally, causing the panel to expand and overlap windows.

Also when the external battery is removed and added back, the added battery percentage label is to the left of the battery icon, instead of above it.

## Actual result

When switching to a vertical panel:
panel1

When swapping the external battery on a system using a vertical panel:
panel2

## Expected result
Would expect a vertical layout in both instances such as this:
panel3

samlane-ma:
## Description

Fixes  #245

Fix issues with battery indicators on vertical panels. (Fixes 

- Fixes the orientation of the battery icons / labels so that on a left/right panel, a system with multiple batteries (i.e. a laptop with an internal and external battery) will display them vertically instead of horizontally. (avoiding an issue which causes the panel to expand beyond its expected size.

- Fixes an issue where if a 2nd battery is added to the system after the panel is running, the percentage label is displayed to the left of the icon instead of above it.

panel3

### Submitter Checklist

- [ x ] Squashed commits with `git rebase -i` (if needed)
- [ x ] Built budgie-desktop and verified that the patch worked (if needed)

### DIFF
--- a/src/applets/status/PowerIndicator.vala
+++ b/src/applets/status/PowerIndicator.vala
@@ -213,6 +213,7 @@ public class PowerIndicator : Gtk.Bin {
 			icon.set_spacing(spacing);
 			icon.set_orientation(orient);
 		}
+		widget.set_orientation(orient);
 	}
 
 	private void update_labels() {
@@ -266,6 +267,7 @@ public class PowerIndicator : Gtk.Bin {
 		icon.label_visible = this.label_visible;
 		devices.insert(object_path, icon);
 		widget.pack_start(icon);
+		change_orientation(widget.get_orientation());
 		toggle_show();
 	}
 


serebit:
The code looks sound to me. However, because I don't think any of us have a system with more than one battery that can be used to replicate this, would you mind recording both what happens on current master and what happens with this commit added, and uploading those recordings here?

samlane-ma:
Gladly!

First example - current master - when the panel is moved to the right, you can see how the icons are side by side and the panel gets expanded.  When I physically remove and reinsert the  external battery, you can see the panel shrink and re-expand, but now with the label next to the battery icon instead of above it:
https://user-images.githubusercontent.com/67085765/210261129-3613bfe9-f28b-44ac-8995-744587a838e5.mp4

Second example - with this commit - when the panel is moved to the right, the batteries are now oriented vertically and the panel doesn't expand.  When I removed the external battery and then reinserted, it reappears with the label above it.
https://user-images.githubusercontent.com/67085765/210261340-b66d953a-39a8-4545-a0b0-22d8a00f504a.mp4

JoshStrobl (approved):
LGTM, thanks and congrats on your first PR!

PR #252, opened 2023-01-15
Update Raven Widgets

serebit:
## Description

Includes the following changes to Raven widgets:

- Clicking on the icons for the Sound Input and Sound Output widgets now mute the audio stream
- Muting the audio stream in Sound widgets disables the slider in the header bar
- "Play Some Music" button was moved to the Media Controls widget
- Made Media Controls player layout significantly more compact
- Marked more text as translatable
- Replaced margins on all widgets with a spacing property on the parent box

These changes were originally in #223 within "Big Sexy", but have been pulled out into the smaller "Little Sexy" for the purposes of reducing scope.

### Submitter Checklist


JoshStrobl:
I'm going to be reverting the symbolic to non-symbolic change. Sorry but I absolutely hate it.

JoshStrobl:
Also as a heads up, I am going to (or have already done) the following changes:

- Reverted the logic that prevented slider change on mute. The Gvc state of muted and volume is different and the behavior is intentional so if you accidentally blast your ears out, you can mute, change the volume, then unmute.
- Reverted the text sizing for devices list and artist info. This should be left up to the themes and it was much harder to read on High DPI + High Resolution.
- Adding some padding / spacing on the media player widget, at least so it'll look alright until I land the new internal theme. The label and artwork alignment is a bit off and the click targets for the buttons need some improvement.

Otherwise once I push this commit then there won't be anything else for me to complain about. I really like being able to mute just by clicking the output button. Thank you for that!

PR #258, opened 2023-01-17
Disable Hibernation button if the system doesn't support it

serebit:
## Description

logind provides a method of detecting if hibernation is enabled on a system. This PR adds a check to the power dialog code that sets the sensitivity of the hibernation button conditional on whether hibernation is supported. Additionally, if hibernation is not supported, a tooltip is added to the button on hover that explains why the button is disabled.

### Submitter Checklist


JoshStrobl (approved):
LGTM, thanks!

PR #260, opened 2023-01-19
Move polkit and power dialogs to libexecdir

serebit:
## Description

Both are internal-use binaries anyway, so they should have been there from the start. Installs to `/usr/libexec/budgie-desktop` on most distributions, `/usr/lib/budgie-desktop` on Arch.

### Submitter Checklist


fossfreedom:
Just need to remove the polkit man help from here https://github.com/BuddiesOfBudgie/budgie-desktop/tree/master/data

JoshStrobl (approved):
LGTM, thanks!

PR #262, opened 2023-01-20
Add new default menu icon

### LINKED ISSUE #10: Don't rely on system icon theme for appmenu icon by default
Putting here as a sort of RFC. Currently, stock budgie uses Adwaita by default due to gschema defaults, and the foot that's displayed as an appmenu icon in Adwaita isn't the best impression for new users. We're not GNOME, and the panel should reflect that.

### Proposal
Use the Papirus appmenu icon or similar as a custom appmenu image by default, leveraging the current functionality to set a custom appmenu icon. If users wish to use their theme's appmenu icon, they can remove the custom image. Alternatively, we could add a switch to Budgie Desktop Settings dedicated to using the appmenu icon from the theme rather than the custom one.

serebit:
## Description

This pull request adds a new symbolic menu icon made from scratch. This new icon is only used if the user enables the new "Use Default Menu Icon" option in the Budgie Menu applet settings, or if the currently selected icon theme is in the following list: `["", "Adwaita", "HighContrast", "hicolor", "default"]`. 

The new icon was fully made from scratch using Inkscape. I've elected to license it under CC0 v1.0, and have mentioned this in the icons folder's README.

**New Icon**
image

**Papirus**
image

**Paper**
image

**hicolor**
image

Resolves #10.

### Submitter Checklist

### DIFF
--- a/src/panel/applets/budgie-menu/com.solus-project.budgie-menu.gschema.xml
+++ b/src/panel/applets/budgie-menu/com.solus-project.budgie-menu.gschema.xml
@@ -14,6 +14,12 @@
       <description>The menu label to display on the menu button</description>
     </key>
 
+    <key type="b" name="use-default-menu-icon">
+      <default>true</default>
+      <summary>Use Default Menu Icon</summary>
+      <description>Use the default icon for the menu button</description>
+    </key>
+
     <key type="s" name="menu-icon">
       <default>'start-here-symbolic'</default>
       <summary>Menu button icon</summary>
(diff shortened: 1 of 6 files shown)


src/panel/applets/budgie-menu/com.solus-project.budgie-menu.gschema.xml
```diff
@@ -14,6 +14,12 @@
       <description>The menu label to display on the menu button</description>
     </key>
 
+    <key type="b" name="use-default-menu-icon">
+      <default>true</default>
```
EbonJaeger, line 18: As discussed on Matrix, let's default this to `false`, and add a check to force it on if any of the following themes are in use: default, Adwaita, hicolor

That way, all the other themes can show their icons by default, while still having a sensible icon for the other ones and still having the option to use ours.

EbonJaeger (changes requested):
Looking good, just the one thing we talked about. :)

EbonJaeger (approved):
Looks good! Thanks!

PR #267, opened 2023-01-23
Return early if a popup shouldn't be shown

### LINKED ISSUE #256: Getting a notification causes screen to flicker when a game is fullscreen
## Description
When a game is fullscreened (in my case, MechWarrior: Online), receiving a desktop notification causes the game to flicker between it and the desktop for a fraction of a second. I'm not sure if this is a regression from #222 or not.

## Budgie version
10.6.4 (git-87e808370836165502ba75349693a1043560fba9)

## Operating system
Solus 4.3 Fortitude

1. Launch a fullscreen game
2. Ask someone to ping you on Matrix (preferably on another device to keep the game focused)
3. Observe screen flicker

## Actual result
Screen flickers

## Expected result
Screen to not flicker

## Additional information
I suspect that the fix for this will be to not create a new window when a popup should not be shown at all, which may necessitate modifying how we handle notifications in Raven.

EbonJaeger:
## Description
Return early in the notification handler if a popup shouldn't be shown. This prevents creating (and showing) a window, even if the popup window isn't visible.

Because of Raven, we can't just return the ID. Instead, dispatch the notification and then close it as Expired right after.

Note that this may only fix a regression from #222, and not from 10.6.

Fixes #256

### Submitter Checklist

### DIFF
--- a/src/daemon/notifications/dbus.vala
+++ b/src/daemon/notifications/dbus.vala
@@ -266,9 +266,25 @@
 							!this.dispatcher.notifications_paused && // notifications aren't paused, e.g. no fullscreen apps
 							(this.popups.size() < MAX_POPUPS_SHOWN || notification.urgency == NotificationPriority.URGENT); // below the number of max popups, or the noti is critical
 
-			// Set to expire immediately if a popup shouldn't be shown.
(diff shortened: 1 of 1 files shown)


serebit:
This fixes #256 on my desktop.

serebit (approved):

PR #268, opened 2023-01-24
Respect OnlyShowIn in Budgie Menu

serebit:
## Description

This adds a specific check to the Budgie Menu that avoids showing an application entry if it should not be shown in Budgie (for example, if the desktop file sets OnlyShowIn and excludes Budgie from the entries). The intent is to hide GNOME-specific software that should not be used on other desktop environments, primarily GNOME Control Center, as Budgie's XDG_CURRENT_DESKTOP is `Budgie:GNOME` for compatibility reasons. 

### Submitter Checklist

### DIFF
--- a/src/appindexer/AppIndex.vala
+++ b/src/appindexer/AppIndex.vala
@@ -311,7 +311,7 @@ namespace Budgie {
 			// We have to make sure to add BCC panel items because they
 			// have NoDisplay set, so this would otherwise exclude them.
 			// Showing/hiding them is handled by the UI layer.
-			bool should_skip = (!app_info.should_show() && !is_control_center_panel) ||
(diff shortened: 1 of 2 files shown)


JoshStrobl:
Absolutely not. This negativity impacts the ability to show applications from other desktop environments or based around them that SHOULD be shown in Budgie Menu, without us having to go around and patch a bunch of applications from MATE, XFCE, XApps from Linux Mint, etc.

serebit:
I think you misunderstand—the only case in which OnlyShowIn doesn't apply for us *at the moment* is for GNOME, and the vast majority of GNOME apps don't set that property. Apps that set OnlyShowIn to something other than GNOME already aren't shown.

PR #270, opened 2023-01-24
Fix notification drop shadow

serebit:
## Description

Fixes a regression from #222 where the drop shadow on notification popups was cut off by the window border.

### Submitter Checklist

### DIFF
--- a/src/daemon/notifications/dbus.vala
+++ b/src/daemon/notifications/dbus.vala
@@ -110,7 +110,7 @@
 		private const string APPLICATION_PREFIX = "/org/gnome/desktop/notifications/application";
 
 		/** Spacing between notification popups */
-		private const int BUFFER_ZONE = 10;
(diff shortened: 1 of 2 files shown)


EbonJaeger (commented):
This does fix the issue, but I feel like the gap between notifications is now too large. Maybe worth playing with the spacing [here](https://github.com/BuddiesOfBudgie/budgie-desktop/blob/main/src/daemon/notifications/dbus.vala#L113) to see if it can be tightened up. inb4 it looks good as-is on high DPI :P

serebit:
Shrank the top margin from 8 to 2, so the total is now 10, and set BUFFER_ZONE to 0. Shadow still fully appears.

EbonJaeger (approved):
Looks good to me! Thanks for fixing this!

JoshStrobl (approved):
LGTM, thanks!

PR #272, opened 2023-01-25
Reorganize data directory

serebit:
## Description

- Move `/docs` to `/data/gtk-docs`
- Move all man pages in `/data` to `/data/man`
- Move logic for each type of data into its own meson.build file in each subdirectory
- Delete unused `/data/glade.css`

### Submitter Checklist


JoshStrobl:
I don't see this really as much of an improvement to be honest. It's excessive splitting for such a small amount of files and documentation IMO belongs in docs.

JoshStrobl (changes requested):
IMO docs should probably be toplevel. Everything else is LGTM. Gonna wait until after 10.7 and its branding to merge this though because there is already a bunch of work for translators double checking things based on moves anyways.

JoshStrobl (approved):
Looks good to me. I'll merge this once I create the 10.7 branch. Thanks!

JoshStrobl:
Needs rebase

JoshStrobl (approved):

PR #273, opened 2023-01-27
Fix missing hide calls and item_clicked emit on overlay menu.

JoshStrobl:
This PR adds function calls to hide Budgie Menu when clicking the Budgie Desktop Settings and System Settings buttons, as well as adding a missing item_clicked signal call in Overlay Menu for closure of Budgie Menu based on the connected function.


serebit (approved):
LGTM!

PR #274, opened 2023-01-27
Fix spacing in buttons in Budgie Menu overlay

serebit:
## Description

Reduces the padding to the left of the icons in the Budgie Menu user dirs overlay.

### Before Patch

image

### After Patch

image

### Submitter Checklist


JoshStrobl (approved):
LGTM!

PR #278, opened 2023-01-29
Bump to 10.7, updated assets and translations.

JoshStrobl:


EbonJaeger (approved):
Looks good to me. I didn't look at all the language files; I figure they're just translation/string updates that I wouldn't understand anyway, lol

PR #283, opened 2023-01-30
Take into account default handlers for relevancy

EbonJaeger:
## Description

Adjust the relevancy score of applications relative to a search term based on if that application is the default handler for the MIME types it supports. This should make it more likely that applications set as default for a content type is shown above other applications.

### Things to Note

We need to make sure that this doesn't cause regressions in searching, where unexpected applications aren't accidentally promoted over others (which would be really ironic given that is exactly what this is meant to combat).

### Submitter Checklist


src/appindexer/RelevancyService.vala
```diff
@@ -132,6 +132,13 @@ namespace Budgie {
 				}
 			}
 
+			// Check if the application is the default handler for it's supported
```
serebit, line 135: *its

serebit (changes requested):
LGTM otherwise.

src/appindexer/RelevancyService.vala
EbonJaeger: Good catch. Fix'd

JoshStrobl (approved):
LGTM, thanks!

PR #285, opened 2023-01-31
Ensure screenshot popover labels match their corresponding keyboard shortcut

fossfreedom:
## Description
This is a case of not cleaning one's glasses during implementation.

The screenshot window "..." shows the current keyboard shortcuts ... except the area and window labels are the wrong way around.  This PR switches around the labels.

### Submitter Checklist


JoshStrobl (approved):
LGTM, thanks!

JoshStrobl:
Built locally, CI failed because it's trying to pull down git submodules from GNOME and their infra is in maintenance / returning 503s. So it's fine, ignoring.

PR #286, opened 2023-02-01
feat: Automatically update Calendar widget when showing Raven.

JoshStrobl:
This PR introduces an update to our Calendar widget that eliminates an unnecessary 30 second timer that existed for the purposes of highlighting the current day. Instead, we just listen for the ExpansionChanged event from Raven and when we know Raven is open, update it then.


EbonJaeger (approved):
LGTM! Have to wait a month to test it though :P

(Yes, I know, changing system time is a thing, but that ruins the joke)

JoshStrobl:
> (Yes, I know, changing system time is a thing, but that ruins the joke)

Don't even have to do that! Change the month via the calendar, close Raven, open it up.

PR #287, opened 2023-02-01
Test if the window of type desktop before taking a window screenshot.

fossfreedom:
## Description
image

Where invoking the screenshot and taking a window screenshot of an empty desktop, the image made is of a transparent (black) desktop displaying icons - screenshot (window) is actually a taking a screenshot of a hidden window.  

In this case we should take a full screenshot rather than of the transparent window.

Tested with nemo that manages the desktop icons - with this PR - a window screenshot no longer just takes a screenshot of a transparent window showing the desktop icons.  Turning off desktop icons support in BDS and making a screenshot (window) screenshots the whole desktop as well.

Note - there should only ever be either zero or one window of window-hint DESKTOP.  I can't think of a scenario why there would be multiple windows of window-type DESKTOP.

### Submitter Checklist


JoshStrobl (approved):
LGTM, thanks!

PR #293, opened 2023-02-06
Plugin chooser: escape description markup symbols

### LINKED ISSUE #292: [Bug]: Plugins with "&" in their descrption fail to show the proper description
### Description

When a plugin has an "&" in their "description=" line in the .plugin file, the description fails to properly show. (set_markup error).

(Markup.escape_text is not used on the description line)

### Budgie version

10.7

### Operating System

Ubuntu Budgie 22.10 / 23.04

1. Open BDS
2. Add a plugin and select one that has an "&" in the description (for example the Budgie Extras Quick Notes applet has "Quick & Easy Notes"
3. Description for that applet does not properly load.

### Actual result

The description of the applet is not updated (it shows the prior selected applet description, in this case the Places applet )

applet

### Expected result

Would expect the applet description to show "Quick & Easy Notes"

### Additional information

I noticed that several of the other lines use Markup.escape_text on the line from the plugin files. I assume it was just an oversight not to have it for the description line as well.

serebit:
## Description

Resolves #292.

### Submitter Checklist


JoshStrobl (approved):
LGTM, thanks!

PR #294, opened 2023-02-06
be more specific about checking for the git commit

eli-schwartz:
Currently, we check if git is installed, and then also check if git can retrieve a commit ID from the current directory. This is a teensy bit slow in case git is installed, but we are building from a tarball. Instead, do an in-process check whether .git exists, and if so then retrieving the git version is mandatory.

This:
- avoids inaccuracies or wasted time if we're not in a git repo
- makes it a fatal error if we're in a git repo but for deeply mysterious reasons decided to uninstall the git program

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -23,9 +23,9 @@ else
     budgie_screensaver_vala_args = ['-D', 'HAVE_GNOME_SCREENSAVER']
 endif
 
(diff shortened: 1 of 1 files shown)


JoshStrobl (approved):
LGTM, thanks!

PR #300, opened 2023-02-11
fix: Inability to capture fullscreen applications like games.

### LINKED ISSUE #298: [Bug]: Budgie-Screenshot does not capture fullscreen games
### Description

When opening the screenshot utility and selecting the screen / window of a game, it will instead display an screenshot of the desktop's wallpaper. Happens on both native and Proton games. Some apps also have this problem but I couldn't reproduce that issue in all of them, and in some extreme cases my Budgie session would crash.

### Budgie version

budgie-desktop 10.7

### Operating System

Fedora 37

1. Open Steam
2. Run a game through Proton's compatibility layer
3. Open Budgie-Screenshot
4.  Take a Window / Monitor screenshot

### Actual result

My desktop's wallpaper gets screenshotted.

### Expected result

An screenshot of the game.

### Additional information

_No response_

JoshStrobl:
Fixes #298. This also cleans up some duplication of logic between screen-type screenshot and area, where screen is just an area of the dimensions of the screen.

Similar to GNOME Shell, we will dynamically disable undirect, before potentially re-enabling it (if it was not explicitly disabled in our settings before the call to screenshot).

Tested in the following games, all running in exclusive fullscreen mode, across several instances of both "screen" and "window" capture to validate that these aren't just one-offs that worked:

- Halo: Reach in MCC
- OpenMW
- Sid Meier's Civilization VI

All the above games were running in exclusive fullscreen modes.

### DIFF
--- a/src/panel/settings/settings_wm.vala
+++ b/src/panel/settings/settings_wm.vala
@@ -81,7 +81,7 @@ namespace Budgie {
 			switch_unredirect = new Gtk.Switch();
 			grid.add_row(new SettingsRow(switch_unredirect,
 				_("Enable unredirection"),
-				_("Enable unredirection which will allow frames to bypass compositing for fullscreen application. This option is for advanced users and recommended to keep enabled. Use this if you are having graphical or performance issues with dedicated GPUs.")
+				_("Enable unredirection which will allow frames to bypass compositing for fullscreen applications. This option is for advanced users and recommended to keep enabled. Use this if you are having graphical or performance issues with dedicated GPUs.")
 			));
 
 			switch_all_windows_tabswitcher = new Gtk.Switch();
--- a/src/panel/settings/settings_wm.vala
+++ b/src/panel/settings/settings_wm.vala
@@ -80,8 +80,8 @@ namespace Budgie {
 			/* Unredirect.. */
 			switch_unredirect = new Gtk.Switch();
 			grid.add_row(new SettingsRow(switch_unredirect,
-				_("Disable unredirection of windows"),
-				_("This option is for advanced users. Use this if you are having graphical or performance issues with dedicated GPUs.")
+				_("Enable unredirection"),
+				_("Enable unredirection which will allow frames to bypass compositing for fullscreen application. This option is for advanced users and recommended to keep enabled. Use this if you are having graphical or performance issues with dedicated GPUs.")
 			));
 
 			switch_all_windows_tabswitcher = new Gtk.Switch();
@@ -114,7 +114,7 @@ namespace Budgie {
 			budgie_wm_settings.bind("pause-notifications-on-fullscreen", pause_notifications, "active", SettingsBindFlags.DEFAULT);
 			budgie_wm_settings.bind("edge-tiling", switch_tiling,  "active", SettingsBindFlags.DEFAULT);
 			budgie_wm_settings.bind("focus-mode", switch_focus, "active", SettingsBindFlags.DEFAULT);
-			budgie_wm_settings.bind("force-unredirect", switch_unredirect, "active", SettingsBindFlags.DEFAULT);
+			budgie_wm_settings.bind("enable-unredirect", switch_unredirect, "active", SettingsBindFlags.DEFAULT);
 			budgie_wm_settings.bind("show-all-windows-tabswitcher", switch_all_windows_tabswitcher, "active", SettingsBindFlags.DEFAULT);
 		}
 	}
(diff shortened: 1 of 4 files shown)


JoshStrobl:
As a sidenote, I feel this logic has always been inverted when it comes to the enabling / disabling, or quite possibly just broken in budgie-wm. It never actually called `disable_unredirect_for_display` if the value was the default `false`, which begs the question then how Mutter was handling it internally. In the case of GNOME Shell, their screenshot functionality will explicitly call disable_unredirect_for_display just before the screenshot, and the enable after the screenshot, whereas ours defaults to false, which meant it never called anything unless it was changed during the session (otherwise the value would be the same), not to mention the "disabling" would turn it "on" which would enable the redirection...not disable it. *head explodes*

Honestly at this point I'd rather just get rid of the option and clean this up, thoughts @serebit,  @EbonJaeger, @fossfreedom? Having us explicitly call to disable during the screenshot then explicitly enable after it's done. Things will just be enabled or whatever Mutter decides in the background otherwise, which is inline which the screenshot functionality in GNOME Shell according to https://gitlab.gnome.org/GNOME/gnome-shell/-/blob/main/src/shell-screenshot.c

JoshStrobl:
Alternatively we can invert our option, so instead of `force-undirect` the option is just `enable-undirect`. The default would be true, instead of false, and we always would disable during the screenshot (which would mean that frame isn't going to the GPU, it's going to Clutter / Mutter for capture) at which point we would re-enable (by default) or do nothing (keep disabled) if the user explicitly has the option disabled to avoid any GPU performance problems. IMO that may be the better option.

JoshStrobl:
The "alternative" proposed option related to switching the option to "enable-unredirect" which will default to on (as should have been the case but wasn't guaranteed to be) has bee committed. I did the same testing as prior, with the games this time being OpenMW and The Elder Scrolls Online. Both fullscreen exclusive and worked as intended. Logic seems saner since we actually restore to unredirection now.

src/panel/settings/settings_wm.vala
```diff
@@ -80,8 +80,8 @@ namespace Budgie {
 			/* Unredirect.. */
 			switch_unredirect = new Gtk.Switch();
 			grid.add_row(new SettingsRow(switch_unredirect,
-				_("Disable unredirection of windows"),
-				_("This option is for advanced users. Use this if you are having graphical or performance issues with dedicated GPUs.")
+				_("Enable unredirection"),
+				_("Enable unredirection which will allow frames to bypass compositing for fullscreen application. This option is for advanced users and recommended to keep enabled. Use this if you are having graphical or performance issues with dedicated GPUs.")
```
serebit, line 84: *for fullscreen applications

src/wm/com.solus-project.budgie.wm.gschema.xml
```diff
@@ -86,10 +86,10 @@
...
 		</key>
 
-		<key type="b" name="force-unredirect">
-			<default>false</default>
-			<summary>Forcibly disable screen unredirection</summary>
-			<description>Disables the screen unredirection which may improve performance with certain drivers.</description>
+		<key type="b" name="enable-unredirect">
+			<default>true</default>
+			<summary>Enable unredirection of fullscreen applications</summary>
+			<description>Enable unredirection which will allow frames to bypass compositing for fullscreen application.</description>
```
serebit, line 92: *for fullscreen applications

serebit (changes requested):
Needs some grammar fixes, otherwise LGTM.

src/wm/com.solus-project.budgie.wm.gschema.xml
JoshStrobl: Fixed in 79d9f2ab34e17015d55e41fe3afddcf84cfc4335

src/panel/settings/settings_wm.vala
JoshStrobl: Fixed in 79d9f2ab34e17015d55e41fe3afddcf84cfc4335

serebit (approved):

fossfreedom:
LGTM! cheers.

EbonJaeger (approved):
Seems straightforward enough to me.

Cherryda:
Absolutely incredible speed! Happily waiting for the changes to make it to the next version available in Fedora testing, whenever that time comes. 

Amazing work, you guys.

JoshStrobl:
> Absolutely incredible speed! Happily waiting for the changes to make it to the next version available in Fedora testing, whenever that time comes.

This'll be in 10.7.1 which I'll release next Saturday :)

PR #301, opened 2023-02-11
Read available languages from LINGUAS file

EbonJaeger:
## Description
Reads the languages from the LINGUAS file instead of having a static list in the meson file.

### Submitter Checklist


JoshStrobl (approved):
YES

eli-schwartz:
> Reads the languages from the LINGUAS file instead of having a static list in the meson file.

Any particular reason why you don't just delete languages handling altogether? If you don't specify it, it will already try to read a file named LINGUAS in the po/ directory.

PR #304, opened 2023-02-16
Implement setting to change sorting in Raven notification view

### LINKED ISSUE #303: [Request] Support changing notifications order
### Budgie version

10.7

### Use-cases

According to the following blog entry, the behavior of notification order was changed.

"Notifications are now sorted oldest to newest."
https://blog.buddiesofbudgie.org/budgie-10-7-released/

But, after evaluating new behavior for a while, it is not suitable when many
 notifications were sent to notification widget. 

Typically, the newest notification is not displayed in
the visible notification widget area because of unread messages,
 need to be scrolling down to show the newest message.

### Proposal

Support both of notification order:

* sorted oldest to newest (default)
* sorted newest to oldest

It may be happy that it is configurable via budgie-desktop-settings.

### References

_No response_

EbonJaeger:
## Description
Adds a setting to Budgie Desktop View to change the sorting order of notifications in Raven Notification View. The two options are from newest to oldest (default), and oldest to newest.

Closes #303 

### Submitter Checklist

### DIFF
--- a/src/panel/settings/settings_raven_settings.vala
+++ b/src/panel/settings/settings_raven_settings.vala
@@ -12,6 +12,7 @@
  namespace Budgie {
 	public class RavenSettingsPage : Budgie.SettingsGrid {
 		private Gtk.ComboBox? raven_position;
+		private Gtk.ComboBox? notification_sort;
 		private Settings raven_settings;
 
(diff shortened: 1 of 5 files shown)


JoshStrobl (approved):
LGTM! Thanks!

PR #305, opened 2023-02-16
Introduce mutter-12 support

JoshStrobl:
## Description

This PR adds support for mutter-12 ABI. Mutter 12 removed get_backend from Meta.Backend, so we need to get the display, and from that the context, through which we can get the backend. These APIs all exist in Mutter 9 and above, so this does not involve extra work with ifdefs and any HAVE_MUTTER_12 (though I carried forward the commented out arg for that just in case.

Tested against Fedora 39 rawhide which has mutter "abi 12" 44.beta, as well as Fedora 38.


fossfreedom (approved):
Tested on ubuntu jammy - screenshot, scaling.  No issues encountered.

Will test with mutter 12 in due course - I spy packages being prepared so suspect will appear in our archive next few days/next week (hopefully)

PR #309, opened 2023-02-19
Tidy-up api's that are no longer part of mutter-12

fossfreedom:
## Description
Tidy up v12 mutter vapi api's that have changed or no longer there

### Submitter Checklist


JoshStrobl (approved):
Thanks for the cleanup!

PR #318, opened 2023-02-24
Relevancy algorithm refinements

### LINKED ISSUE #316: [Bug] Less useful application search results since updating
### Description

Hey, I'm not entirely sure how to describe this, but the search results when searching for applications seem to have become less... useful? since updating. 

### Budgie version

10.7.1

### Operating System

Arch Linux

Just as an repro example:
1. search "vscode"
2. search "ter"

### Actual result

1. The Itch launcher is the first result. 
2. "LibreOffice Writer" is the first result. 

### Expected result

1. "Visual Studio Code" is the first result. 
2. "Terminal" is the first result. 
### Additional information

_No response_

EbonJaeger:
## Description
This reworks the searching algorithm to use a [fuzzy Bitap](https://en.wikipedia.org/wiki/Bitap_algorithm) text matching algorithm for finding possibly relevant applications by name. This differs from the previous implementation (which only used basic Levenshtein distance across the entire strings) mainly in that it sees if the text to search contains any substring matching the search term.

To maybe make that easier to understand, searching for Budgie Control Center with the term "control" with the old algorithm would consider Budgie Control Center less relevant because it only calculates the distance of the entire string, whereas this algorithm ranks it more relevant because it does substring matching.

There are a few other smaller changes made in narrowing down the results:
- If an exact match is found, worsen the score of matches where the term is not the start of the application name
- If an exact match is found, do no further checking since the score is already the best it can be
- Ensure that only applications with a positive relevancy score are shown
- Don't improve the score of applications if the search term appears in the application's metadata if the term is not the beginning of a word
  - For example, previously the score would be improved if the term was "ter" and a keyword was "battery". Now, it would not match "battery" but it would match "terminal", because terminal starts with the search term. This helps reduce false positives.
- If there are multiple perfect scores, sort them in the view alphebetically

Closes #316 

### Submitter Checklist

### DIFF
--- a/src/appindexer/RelevancyService.vala
+++ b/src/appindexer/RelevancyService.vala
@@ -174,43 +174,47 @@ namespace Budgie {
 		 *
 		 * @param text The text to compare to.
 		 * @param pattern The pattern to match against.
-		 * @param k The maximum distance between the strings to still be considered equal.
+		 * @param max_distance The maximum distance between the strings to still be considered equal.
 		 */
-		private int get_fuzzy_score(string text, string pattern, int k) {
+		private int get_fuzzy_score(string text, string pattern, int max_distance) {
 			var result = -1;
-			var m = pattern.length;
-			int[] R;
+			var pattern_length = pattern.length;
+			int[] bit_array;
 			int[] pattern_mask = new int[128];
 			int i, d;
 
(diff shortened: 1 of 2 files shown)


JoshStrobl (approved):
LGTM, thanks!

PR #320, opened 2023-03-03
feat: ensure Budgie Menu application lists height allocation accounts for GTK scaling factor

JoshStrobl:
This ensures all parts of Budgie Menu are visible when used on resolutions such as 1080p with 200% scaling, rather than parts of it being cut off.


serebit (approved):
LGTM, thanks!

PR #322, opened 2023-03-08
Limit total number of notifications within Raven

Jacob-Vlijm:
## Description
This pull request limits the number of notifications per group to 12. There was a report that notifications would accumulate if user would not attend notifications on raven, eventually leading to a non-responsive desktop with thousands of notifications https://discourse.ubuntubudgie.org/t/raven-issue-unread-notifications/6603.
Currently, the number of notifications per group is hard-coded to twelve, which is not the most elegant way to go. Preferably, this would be set in a gschema, and possibly in BDS. Currently, budgie-raven settings are defined in class Raven, and not available in class NotificationGroup, so a namespace revision could be considered. Request for feedback on that.

### Submitter Checklist


EbonJaeger (dismissed):
I'm not sure how I feel about arbitrarily just not showing notifications in Raven if there are more than *x* amount of them. I recognize it presents a problem, but I can't help but feel that there must be better ways to address it; I recall @JoshStrobl in the past saying he had some ideas. I would also like to know how this is handled on other platforms (Windows, macOS, other DE's, etc).

If we do decide to go this route, however, I think it would be way simplier and easier to check the number of current notifications in the group whenever a new notification is added, and then just not add it if it would take us over the limit. There's no need to do this in the count update and then iterate over all the notifications in the group every time, is there?

Jacob-Vlijm:
The ultimate way IMO would be to keep the notifications, but limit the presentation in widgets, which is most likely/certainly the issue. Then with a "more..." button or something have the next n show.
That would be quite an extensive edit though, while I think the issue is kind of urgent. Without knowing it was this issue, had my desktop choke a few times over this.
Just keep older ones, skipping new ones would lead to odd situations IMO.

EbonJaeger:
> Just keep older ones, skipping new ones would lead to odd situations IMO.
> Think we are not iterating over all of the notifications, just the ones to be deleted, which is most of the time 1.

Ah yes, you're right. And there really isn't a good way to get either the first or last element of the list that would be better than doing another sort...

> That would be quite an extensive edit though, while I think the issue is kind of urgent.

It's been this way for a number of years now, I think the urgency has long since passed. I'd rather see it fixed right (preferably the first time) instead of doing a half-measure. The trick is figuring out what the "right" way is.

EbonJaeger (commented):
> Just keep older ones, skipping new ones would lead to odd situations IMO.
> Think we are not iterating over all of the notifications, just the ones to be deleted, which is most of the time 1.

Ah yes, you're right. And there really isn't a good way to get either the first or last element of the list that would be better than doing another sort...

> That would be quite an extensive edit though, while I think the issue is kind of urgent.

It's been this way for a number of years now, I think the urgency has long since passed. I'd rather see it fixed right (preferably the first time) instead of doing a half-measure. The trick is figuring out what the "right" way is.

I do like the idea of hiding them behind a "More" button, and actually thought about that this morning. Someone would have to make sure that it actually solves (or helps solve) the problem.

EbonJaeger:
Sorry for the extra notifications while I fight with the GitHub review system.

fossfreedom:
> I do like the idea of hiding them behind a "More" button, and actually thought about that this morning. Someone would have to make sure that it actually solves (or helps solve) the problem.

For the curious here - if we are holding - say - 6000 notifications of lets say 100chars each - thats 600Kb of memory - so relatively small in the scheme of things.  So how are we running out of memory without end-users noticing until it hits?  Are we poking each notification straight into raven when its received?  I'm assuming here the guilty party here is how much memory the GUI part is taking?

Thinking of sane defaults here - lets say Discord is the guilty "notification" generator - is someone really going to click "more" approx 500-600 times to get to the bottom of the list when we display say 10 notifications?  I'm doubtful anyone is interested in anything more than the latest notifications.  So to me its more of a question of how much the maximum should be displayed in a group before end-users become uninterested in what they are reviewing!

JoshStrobl:
The problem here is Gtk3's ListBox does not perform virtualization of list items. It does in GTK4, but in 3.x it will render all the items at all times, even if they are in view. Capping the amount of items only happens to work around the problem because it reduces the amount of items rendered in one go and even that won't necessarily work if the individual notifications have a large amount of text.

Hiding a subset of items behind a revealer + incrementally allowing more items to render within a given notification group might work, but that depends on how GTK handles renders of revealer children.

Jacob-Vlijm:
I said:

> The ultimate way IMO would be to keep the
> notifications, but limit the representation in
> widgets, which is most likely/certainly the
> issue. Then with a "more..." button or something have the next n show.

The question is however what it would add to the current pr. It would make things far more complicated, but still you'd need to set s limit to the stored notifications *at some point*, and it is highly unlikely anyone is interested in a notification hundreds of notifications back. Since the most expensive part is the widget itself, I don't think a notification with a bit more text is an issue.
If you'd ask me it's also more a question how many notifications per group should be kept.

Jacob-Vlijm:
Did a test setup. Not on the latest and greatest box, but still one that generously meets the requirements for a modern Linux desktop.

This is the setup:

- Have a script spit out notifications every 2 sec.
- Starting with zero notifications, panel on ~90mb memory
- Run the script until ~6000 notifications are done, intermediate checks on how raven responds

This is what happens:
- After ~800 notifications, raven "stops being pretty": a noticeable decrease in responsiveness, at times less fluent animation
- After 1100, at times, raven animation is now a 2-step thing.
- After 1400, appearing is sluggish, animation is often no more, just bam, there it is, behavior is ugly.
- After 2500, effectively broken, it does work, but takes too long to be of real use.
- After 5000 notifications (calling raven from idle) ... 11 seconds before it appears.

IMO we can't expect everyone to use the DE like we planned. People not using raven won't be an exception.

JoshStrobl:
> The question is however what it would add to the current pr. It would make things far more complicated, but still you'd need to set a limit to the stored notifications at some point, and it is highly unlikely anyone is interested in a notification hundreds of notifications back.

My line of thinking for a solution, at least until we aren't limited by GTK3 (well, GTK at all), is below. I invite other solutions / feedback.

1. A revealer that allows for displaying additional notifications over a given limit for the listing outside of the revealer. This would keep the main notifications view more readable.
2. Global and per Notification Group caps. As you mentioned and I agree, it isn't likely someone will be reading email notifications or Discord messages from a long time ago or care about the 200th message from a Discord server, and if they are getting them in short sessions then I don't think it is unreasonable to expect that user to curate their application-specific notifications via its respective settings anyways since they would otherwise be getting visual and possibly auditory noise. The gap per notification group should be `global cap / number of notification groups`, so we don't punish added notification groups with a small notification group cap, rather balance over time. As we clear notification groups, the numbers will balance out and additional messages can be accepted evenly across the remaining notification groups (so if the cap for the group was 50 and we had 5 remaining groups, each one could then store 10 more).
3. While the visual sorting order for notifications can be anything, the actual order should be newest to oldest, with oldest being last and popped off the notification group revealer. If we get a notification that comes in, we are still under the global + notification group cap, but over the "show outside revealer" cap, we shift the oldest from the list into the revealer.

This would be more work but I think that would be acceptable seeing as this "problem" has been around for something like 5-6 years at least and there isn't any particular rush to get a less-than-thorough solution in. Whether it gets fixed in 1 day or months won't matter much to most users.

EbonJaeger:
> We should look much closer at the current limits for a reasonable desktop setup - is it 100, 500, 1000 etc for a global cap?

@Jacob-Vlijm didn't say what computing resources were available to the test setup, but it would at least indicate that the limit should be under 800 total notifications at least, and likely actually lower than that. 

Having the per-group limit being dynamic is a neat idea, I like that.

As @fossfreedom said, adding a revealer to show more notifications in a group should be a separate PR. Heck, with the limits, implemented, we may find it no longer necessary.

JoshStrobl:
Yea in agreement that it can be separate PRs if we find the revealer necessary.

fossfreedom:
@EbonJaeger from your expert knowledge in this area one of the questions made by Jacob initially was how we handle these "fixed-size" numbers - should they be hard-coded?  Should they be in a gschema? If the latter something new?  Changing namespaces to reuse the raven schema?

EbonJaeger:
Given that the limit for notifications might be in the hundreds, I don't think it makes sense to expose that as a setting to users directly in the settings app. Putting it in a gschema might be a good idea, and makes it easier to test different values, but a `const int MAX_NOTIFICATIONS = 500;` would work just as well.

If it is in a gschema, then I think it should be in Raven's under the `/com/solus-project/budgie-raven/` path.

Jacob-Vlijm:
Have also been playing around with some ideas. A dynamically changing number of notifications _per group_ seems obvious. At the same time, from a user point of view, I wouldn't mind a limited number, but still would want to know what that number is -by default-. Dynamically changing I would therefore find confusing.
Came to the idea that maybe we should set a max per group that we find acceptable (somewhere between 10-25) , then -only- if the total number of notifications exceeds the max we find "safe" (500?) , set the groups on ration (500/n-groups).
Thinking of a group max of (even) 25, it would take up to 20 applications with 25 notifications before we'd need to start the "safety procedure", which seems very unlikely to ever happen.

EbonJaeger:
> Came to the idea that maybe we should set a max per group that we find acceptable (somewhere between 10-25) , then -only- if the total number of notifications exceeds the max we find "safe" (500?) , set the groups on ration (500/n-groups).
> Thinking of a group max of (even) 25, it would take up to 20 applications with 25 notifications before we'd need to start the "safety procedure", which seems very unlikely to ever happen.

This seems good to me. I think you're right in that a dynamically changing limit might be confusing to people, especially since there's no real way to convey what's going on.

25 per group with a 500 total max seems like a good balance between not showing enough and showing too many notifications, and as you said, it would still take a *lot* of applications generating a lot of notifications to reach the limit. I doubt it would be reached in practice, which is good. 

> From a technical point of view, it would save a lot of calculation, each and every time the number of groups changes (and the 500 is reached).

Definitely in favor of anything that saves on computation 😛

Jacob-Vlijm:
Edited the PR. Now  DEFAULT_MAX_NOTIFICATIONS_PERGROUP is set to 25, TOTAL_MAX_NOTIFICATIONS to 500.  If the latter is exceeded, the code reduces max_per_group. 
What we forgot in the discussion so far is that in case the number of groups is reduced, max_per_group probably could be optimized (back) to a higher number, as long as it is lower than DEFAULT_MAX_NOTIFICATIONS_PERGROUP. The code now takes care of that too.

Given the fact that it does, we could consider a higher default. If we set for example 50 as a default max, it would take 10 applications with 50 notifications each before the code needs to intervene. Still not very likely to happen ever.

src/raven/notifications_group.vala
```diff
@@ -135,18 +137,62 @@ namespace Budgie {
...
+		 * too many notifications will choke raven and the desktop, so let's set a limit;
+		 * keep the latest n-notifications of current group, delete older ones
+		 */
+		public void limit_notifications () {
+
+			GLib.List<uint32> currnotifs = notifications.get_keys();
+			CompareFunc<int> intcmp = (a, b) => {
+				return (int) (a > b) - (int) (a < b);
+			};
+			currnotifs.sort(intcmp);
```
EbonJaeger, line 156:
```suggestion
			currnotifs.sort((a, b) => {
				return (int) (a > b) - (int) (a < b);
			});
```

src/raven/notifications_group.vala
```diff
@@ -135,18 +137,62 @@ namespace Budgie {
...
+			uint n_currnotifs = currnotifs.length();
+			/**
+			 * no need to reduce if the current number of notifications is below our threshold
+			 * and we shouldn't attempt to set a negative uint
+			 */
+			if (n_currnotifs <= tokeep) {
+				return;
+			}
+			uint n_remove = n_currnotifs - tokeep;
+			print(@"inside: count = $count, n_remove is $n_remove\n");
```
EbonJaeger, line 166:
```suggestion
			debug("inside: count = %d, n_remove is %d", count, n_remove);
```

src/raven/notifications_group.vala
```diff
@@ -135,18 +137,62 @@ namespace Budgie {
...
+			int count = 0;
+			foreach (uint n in currnotifs) {
+				if (count < n_remove) {
+					remove_notification(n);
+				}
+				else {
+					break;
+				}
+				count += 1;
```
EbonJaeger, line 175:
```suggestion
				count++;
```

src/raven/notifications_group.vala
```diff
@@ -135,18 +137,62 @@ namespace Budgie {
...
+			}
+		}
+
+		/**
+		 * if the total number of notifications exceeds threshold, NotificationsView will
+		 * update the max number per group ('tokeep')
+		 */
+		public void set_group_maxnotifications(uint keep) {
+			tokeep = keep;
+			print(@"newtokeep set in group: $tokeep\n");
```
EbonJaeger, line 185:
```suggestion
			debug("newtokeep set in group: %d", tokeep);
```

src/raven/notifications_view.vala
```diff
@@ -318,12 +321,46 @@ namespace Budgie {
...
+
+		void check_notification_allocation(uint len) {
+			uint n_groups = notification_groups.length;
+			uint newmax = max_per_group;
+			bool trim = false;
+			if (n_groups == 0) {
+				/* if no notifications left, set to DEFAULT_MAX_PERGROUP */
+				newmax = DEFAULT_MAX_PERGROUP;
+			}
+			else if (len > TOTAL_MAX_NOTIFICATIONS) {
```
EbonJaeger, line 344: Let's keep chained else/if statements on the same line:
```suggestion
			} else if (len > TOTAL_MAX_NOTIFICATIONS) {
```

src/raven/notifications_view.vala
```diff
@@ -318,12 +321,46 @@ namespace Budgie {
...
+				/* if totalmax is exceeded, reduce max per group */
+				newmax = (uint)((float)TOTAL_MAX_NOTIFICATIONS/(float)n_groups);
+				trim = true;
+			}
+			else if (
```
EbonJaeger, line 349: Same as above

src/raven/notifications_view.vala
```diff
@@ -318,12 +321,46 @@ namespace Budgie {
...
+				max_per_group < DEFAULT_MAX_PERGROUP &&
+				(n_groups * (max_per_group + 1)) <= TOTAL_MAX_NOTIFICATIONS
+			) {
+				/* if we previously reduced group size but now can add at least one back */
+				newmax = (uint)((float)TOTAL_MAX_NOTIFICATIONS/(float)n_groups);
+			}
+			if (newmax != max_per_group) {
```
EbonJaeger, line 356: Same as above

EbonJaeger (changes requested):
Made some minor stylistic suggestions to fit with the rest of the codebase.

JoshStrobl:
> What we forgot in the discussion so far is that in case the number of groups is reduced, max_per_group probably could be optimized (back) to a higher number

This wasn't forgotten. As I suggested:

>  As we clear notification groups, the numbers will balance out and additional messages can be accepted evenly across the remaining notification groups (so if the cap for the group was 50 and we had 5 remaining groups, each one could then store 10 more).

https://github.com/BuddiesOfBudgie/budgie-desktop/pull/322#issuecomment-1462117905

src/raven/notifications_group.vala
Jacob-Vlijm: This was a line, just for checking if all went well, can be removed.

src/raven/notifications_group.vala
Jacob-Vlijm: This was a line, just for checking if all went well, can be removed.

Jacob-Vlijm: This was a line, just for checking if all went well, can be removed.

Jacob-Vlijm:
I have no objections (obviously) to any styling suggestion to make it match the BoB styling conventions. Please review.

src/raven/notifications_group.vala
```diff
@@ -135,18 +137,61 @@ namespace Budgie {
...
+			if (n_currnotifs <= tokeep) {
+				return;
+			}
+			uint n_remove = n_currnotifs - tokeep;
+			print(@"inside: count = $count, n_remove is $n_remove\n");
+			int count = 0;
+			foreach (uint n in currnotifs) {
+				if (count < n_remove) {
+					remove_notification(n);
+				}
```
fossfreedom, line 170: should be `} else {` style convention

src/raven/notifications_group.vala
```diff
@@ -135,18 +137,61 @@ namespace Budgie {
...
+				else {
```
fossfreedom, line 171: should be } else { style convention

src/raven/notifications_view.vala
```diff
@@ -318,12 +321,45 @@ namespace Budgie {
...
+			uint newmax = max_per_group;
+			bool trim = false;
+			if (n_groups == 0) {
+				/* if no notifications left, set to DEFAULT_MAX_PERGROUP */
+				newmax = DEFAULT_MAX_PERGROUP;
+			} else if (len > TOTAL_MAX_NOTIFICATIONS) {
+				/* if totalmax is exceeded, reduce max per group */
+				newmax = (uint)((float)TOTAL_MAX_NOTIFICATIONS/(float)n_groups);
+				trim = true;
+			}
```
fossfreedom, line 347: should be } else if { style convention

JoshStrobl (changes requested):
Given only items raised are stylistic, I'd say this is good to go once they are fixed. Nice work!

Jacob-Vlijm:
Edited suggestions.

JoshStrobl (approved):
LGTM, thanks!

EbonJaeger (approved):
Also LGTM! Thanks!

PR #329, opened 2023-03-14
Fix panel growing memory consumption over increasing numbers of notif…

### LINKED ISSUE #327: [Bug]  Raven seems to have a memory leak
### Description

During testing a follow up on [this PR](https://github.com/BuddiesOfBudgie/budgie-desktop/pull/322), about Raven notifications section, another issue occurred.  The latest tested fix to control the behaviour was tested overnight, pestering Raven with large numbers of notifications (>10.000). 
The good news is that running the fix, Raven stayed crispy and completely responsive.

What i did not expect is that memory consumption increased, despite the limited number of kept notifications, from ~80MB to almost 600MB. Dismissing all notifications did not make the memory available unfortunately, which means memory consumption is increasing until we either restart the panel or restart at all.

To make sure the issue was not the result of the edits, made in the PR, I repeated the test with the original budgie-desktop with the exact same result.

### Budgie version

10.7.1

### Operating System

Ubuntu Budgie 23.04

1. Restart the panel
2. Check memory usage of budgie-panel
3. Run a large number of notifications (a few hundred will already clearly show the issue
4. Check memory again, note the increase
5. Dismiss all notifications

### Actual result

Memory consumption does not decrease on dismiss of the notifications.

### Expected result

Memory consumption decrease to initial state.

### Additional information

Just did the same test on 22.04/budgie 10.5.1, again with the exact same result.

Jacob-Vlijm:
…ications.

## Description
This pull request fixes #327. Tests were done up to over 20.000 notifications from various resources, no noticeable memory increase occurred any more.

### Submitter Checklist

### DIFF
--- a/src/raven/notifications_group.vala
+++ b/src/raven/notifications_group.vala
@@ -113,9 +113,10 @@ namespace Budgie {
 		 * dismiss_all is responsible for dismissing all notifications
 		 */
 		public void dismiss_all() {
-			notifications.foreach_steal((id, notification) => {
-				list.remove(notification.get_parent());
-				notification.destroy();
+			notifications.foreach_remove((id, notification) => {
+				var parent = notification.get_parent();
+				list.remove(parent);
+				parent.destroy();
 				dismissed_notification(id);
 				return true;
 			});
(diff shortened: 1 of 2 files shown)


EbonJaeger (approved):
LGTM! Thanks!

JoshStrobl (approved):
LGTM, thanks!

PR #330, opened 2023-03-16
Use correct C header for meta_keybindings_set_custom_handler

fweimer-rh:
At least in Mutter 12, the function is declared in <meta/keybindings.h>, and not in <meta/main.h>.

This is necessary to avoid a build failure with future compilers which do not accept implicit function declarations by default.

Related to:

* https://fedoraproject.org/wiki/Changes/PortingToModernC
* https://fedoraproject.org/wiki/Toolchain/PortingToModernC

There are further compiler failures after this, but they look more like a Vala compiler bug.


JoshStrobl (approved):
LGTM, thanks!

Issue #328, opened 2023-03-13
New Status Notifier applet failed to register host


EbonJaeger:
### Description

Upon booting the system, I noticed I only had a tray item for Blueman. Tray items for nm-applet and the firewall were not present, and opening Element did not show a tray item for Element. Looking in the journal revealed that it failed to register a host:
```
Mar 13 16:16:53 maddock-desktop-endeavour budgie-panel[1052]: TrayApplet.vala:128: Unable to fetch existing status notifier items: GDBus.Error:org.freedesktop.DBus.Error.UnknownMethod: Object does not exist at path “/org/freedesktop/StatusNotifierWatcher”
Mar 13 16:16:53 maddock-desktop-endeavour budgie-panel[1052]: TrayApplet.vala:152: Failed to register Status Notifier host: GDBus.Error:org.freedesktop.DBus.Error.UnknownMethod: Object does not exist at path “/org/freedesktop/StatusNotifierWatcher”
```

No other errors appear in the logs that I can see.

### Budgie version

budgie-desktop 10.7.1 (git-76d016819853297a595430f4a98be3cfb0c84d3c)

### Operating System

Endeavour OS Cassini Neo

### Steps to reproduce the issue

1. Boot system
2. See missing tray items
3. ?????
4. Profit

### Actual result

No tray items for nm-applet or firewall.

### Expected result

Tray items for nm-applet and firewall.

### Additional information

Here is the full journalctl log for this boot.

[status_tray.log](https://github.com/BuddiesOfBudgie/budgie-desktop/files/10961672/status_tray.log)

PR #333, opened 2023-03-19
Add timeout to watcher init in tray applet

serebit:
## Description

#328 is caused by the tray applet attempting to fetch the watcher's DBus object before it's actually been registered to the bus, which is rare (as it has to happen IMMEDIATELY after the name is owned). This adds a 500ms timeout after the name is found, so the object has plenty of time to register itself.

### Submitter Checklist


JoshStrobl (approved):
LGTM, thanks!

PR #339, opened 2023-03-25
Register a daemon-side SN host

serebit:
## Description

This allows the panel to wait to attach to the host, while also ensuring that applications that start before the applet initializes can still register items. The wait before applet initialization has also been reduced from 500ms to 100ms.

### Submitter Checklist


JoshStrobl (approved):
LGTM, thanks!

PR #352, opened 2023-04-12
power-indicator: add controls for power mode via power-profiles-daemon

### LINKED ISSUE #342: [Request] Allow switching between power modes when clicking the battery button on Status Indicator applet
### Budgie version

10.6.1

### Use-cases

On Windows and KDE Plasma, users can quickly switch between high performance / balanced / power saver modes when clicking on the battery icon in the system tray. This is a very convenient feature and I would greatly appreciate if it was added to Budgie.

### Proposal

The exact implementation could vary, but here's an idea, based on what I recall KDE Plasma's implementation behaves like:
- Create a slider with 3 stops, with a leaf icon on the left and a lightning bolt on the right. 
- Above the slider, write "Power Mode: <selected mode>". 
- Changing the slider changes the above text, intuitively with power saving on the left and performance on the right. 
- If #341 is implemented, then changing the slider should  ONLY change the power mode for the current power state (e.g. if I'm on battery and switch to power saver, then it should ONLY set me to power saver on battery, and when I plug in, it should change back to whatever I had set for "Plugged In". 

Example use case to clarify the last point:
Original setting: Plugged In => Balanced, On Battery => Balanced
Current power state: on battery
Click battery icon -> Change slider to Power Saving. Power mode changes to Power Saving.
Plug laptop in => Power mode changes back to Balanced.
Click battery icon => Change slider from Balanced to Performance. Power mode changes to Performance.
Unplug battery => Power mode changes back to Power Saving.

### References

#341

justchen1369:
## Description
Closes #342

### Submitter Checklist

### DIFF
--- a/src/panel/applets/status/PowerIndicator.vala
+++ b/src/panel/applets/status/PowerIndicator.vala
@@ -169,6 +169,7 @@ public class PowerProfilesSelector : Gtk.Box {
 
 	public PowerProfilesSelector(PowerProfilesDBus profiles_proxy) {
 		orientation = Gtk.Orientation.VERTICAL;
+		spacing = 6;
 
 		var profiles = new GenericSet<string>(str_hash, str_equal);
 
@@ -194,11 +195,11 @@ public class PowerProfilesSelector : Gtk.Box {
 		pack_start(sep, false, false, 1);
 
 		var header = new Gtk.Label("");
-		header.set_markup("<b>%s</b>".printf(_("Power Mode")));
-		header.set_halign(Gtk.Align.CENTER);
+		header.set_markup("<b>%s</b>".printf(_("Performance Mode")));
+		header.set_halign(Gtk.Align.START);
 		pack_start(header);
 
-		var power_profiles_radio_box = new Gtk.Box(Gtk.Orientation.HORIZONTAL, 1);
+		var power_profiles_radio_box = new Gtk.Box(Gtk.Orientation.VERTICAL, 2);
 
 		Gtk.RadioButton radio_group = null;
 
--- a/src/panel/applets/status/PowerIndicator.vala
+++ b/src/panel/applets/status/PowerIndicator.vala
@@ -146,11 +146,117 @@ public class BatteryIcon : Gtk.Box {
 	}
 }
 
+const string POWER_PROFILES_DBUS_NAME = "net.hadess.PowerProfiles";
+const string POWER_PROFILES_DBUS_OBJECT_NAME = "/net/hadess/PowerProfiles";
+
+[DBus (name = "net.hadess.PowerProfiles")]
+public interface PowerProfilesDBus : Object {
+	public abstract HashTable<string, Variant>[] profiles { owned get; }
+	public abstract string active_profile { owned get; set; }
+}
+
+public class PowerProfilesOption : Gtk.RadioButton {
+	public PowerProfilesOption(PowerProfilesDBus profiles_proxy, string profile_name, string display_name) {
+		label = display_name;
+		
+		this.toggled.connect(() => {
+			if (this.get_active()) {
+				profiles_proxy.active_profile = profile_name;
+			}
+		});
+	}
+
+}
+
+public class PowerProfilesSelector : Gtk.Box {
+	// Power profile toggles
+	private PowerProfilesOption radio_power_save;
+	private PowerProfilesOption radio_power_balanced;
+	private PowerProfilesOption radio_power_performance;
+
+	public PowerProfilesSelector(PowerProfilesDBus profiles_proxy) {
+		orientation = Gtk.Orientation.VERTICAL;
+		
+		var profiles = new GenericSet<string>(str_hash, str_equal);
+
+		foreach (HashTable<string, Variant> profile in profiles_proxy.profiles) {
+			var profile_value = profile.get("Profile");
+
+			if (!profile_value.is_of_type(VariantType.STRING)) continue;
+			
+			profiles.add(profile_value.get_string());
+		}
+		
+		// need at least two options for it to be meaningful
+		if (profiles.length < 2) return;
+
+		// initialize state
+		on_active_profile_changed(profiles_proxy.active_profile);
+
+		((DBusProxy) profiles_proxy).g_properties_changed.connect(() => {
+			on_active_profile_changed(profiles_proxy.active_profile);
+		});
+
+		var sep = new Gtk.Separator(Gtk.Orientation.HORIZONTAL);
+		pack_start(sep, false, false, 1);
+
+		var header = new Gtk.Label("");
+		header.set_markup("<b>%s</b>".printf(_("Power Mode")));
+		header.set_halign(Gtk.Align.CENTER);
+		pack_start(header);
+
+		var power_profiles_radio_box = new Gtk.Box(Gtk.Orientation.HORIZONTAL, 1);
+
+		Gtk.RadioButton radio_group = null;
+
+		if (profiles.contains("power-saver")) {
+			radio_power_save = new PowerProfilesOption(profiles_proxy, "power-saver", _("Power Saver"));
+			radio_power_save.join_group(radio_group);
+			radio_group = radio_power_save;
+			power_profiles_radio_box.pack_start(radio_power_save, false, false, 1);
+		}
+
+		if (profiles.contains("balanced")) {
+			radio_power_balanced = new PowerProfilesOption(profiles_proxy, "balanced", _("Balanced"));
+			radio_power_balanced.join_group(radio_group);
+			radio_group = radio_power_balanced;
+			power_profiles_radio_box.pack_start(radio_power_balanced, false, false, 1);
+		}
+
+		if (profiles.contains("performance")) {
+			radio_power_performance = new PowerProfilesOption(profiles_proxy, "performance", _("Performance"));
+			radio_power_performance.join_group(radio_group);
+			radio_group = radio_power_performance;
+			power_profiles_radio_box.pack_start(radio_power_performance, false, false, 1);
+		}
+
+		pack_start(power_profiles_radio_box);
+	}
+
+	void on_active_profile_changed(string active_profile) {
+		switch(active_profile) {
+			case "power-saver":
+				radio_power_save.set_active(true);
+				break;
+			case "balanced":
+				radio_power_balanced.set_active(true);
+				break;
+			case "performance":
+				radio_power_performance.set_active(true);
+				break;
+		}
+	}
+}
+
 public class PowerIndicator : Gtk.Bin {
 	/** Widget containing battery icons to display */
 	public Gtk.EventBox? ebox = null;
 	public Budgie.Popover? popover = null;
 	private Gtk.Box widget = null;
+	private Gtk.Box box = null;
+
+	private PowerProfilesDBus profiles_proxy;
+	private PowerProfilesSelector power_profiles_selector;
 
 	/** Our upower client */
 	public Up.Client client { protected set; public get; }
@@ -170,7 +276,7 @@ public class PowerIndicator : Gtk.Bin {
 		ebox.add(widget);
 
 		popover = new Budgie.Popover(ebox);
-		var box = new Gtk.Box(Gtk.Orientation.VERTICAL, 1);
+		box = new Gtk.Box(Gtk.Orientation.VERTICAL, 1);
 		box.border_width = 6;
 		popover.add(box);
 
@@ -196,10 +302,43 @@ public class PowerIndicator : Gtk.Bin {
 
 		client = new Up.Client();
 
+		Bus.watch_name(BusType.SYSTEM, POWER_PROFILES_DBUS_NAME, BusNameWatcherFlags.NONE, has_power_profiles, lost_power_profiles);
+
 		this.sync_devices();
 		client.device_added.connect(this.on_device_added);
 		client.device_removed.connect(this.on_device_removed);
 		toggle_show();
+	}
+
+	void has_power_profiles() {
+		if (profiles_proxy != null) {
+			create_power_profiles_options();
+			return;
+		}
+		
+		Bus.get_proxy.begin<PowerProfilesDBus>(BusType.SYSTEM, POWER_PROFILES_DBUS_NAME, POWER_PROFILES_DBUS_OBJECT_NAME, 0, null, on_proxy_get);
+	}
+	
+	void lost_power_profiles() {
+		power_profiles_selector.destroy();
+	}
+
+	void on_proxy_get(Object? o, AsyncResult? res) {
+		try {
+			profiles_proxy = Bus.get_proxy.end(res);
+
+			if (profiles_proxy.active_profile != null) 
+				create_power_profiles_options();
+
+		} catch (Error e) {
+			warning("unable to connect to net.hadess.PowerProfiles: %s", e.message);
+		}
+	}
+
+	void create_power_profiles_options() {
+		power_profiles_selector = new PowerProfilesSelector(profiles_proxy);
+		box.pack_start(power_profiles_selector);
+		box.show_all();
 	}
 
 	public void change_orientation(Gtk.Orientation orient) {
--- a/src/panel/applets/status/PowerIndicator.vala
+++ b/src/panel/applets/status/PowerIndicator.vala
@@ -57,9 +57,7 @@ public class BatteryIcon : Gtk.Box {
 	}
 
 	private void on_battery_change(Object o, ParamSpec sp) {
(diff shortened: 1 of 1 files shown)


src/panel/applets/status/PowerIndicator.vala
```diff
@@ -192,6 +204,53 @@ public class PowerIndicator : Gtk.Bin {
 		button.clicked.connect(open_power_settings);
 		button.get_child().set_halign(Gtk.Align.START);
 		box.pack_start(button, false, false, 0);
+
+		power_profiles = new PowerProfiles();
+		
+		power_profiles.proxy_connected.connect(() => {
```
fossfreedom, line 210: If PowerProfiles is invalid for any reason e.g. bus issues, powerprofiles are not applicable to the desktop the proxy_connected would be invalid?

Probably should test this - if there is a failure then powerprofiles in the indicator should probably be hidden

justchen1369: proxy_connected is a signal emitted by PowerProfiles.vala- if it doesn't get the bus for whatever reason, the callback simply shouldn't ever be called. There may be a better way to structure it to make that clear, though.

src/panel/applets/status/PowerProfiles.vala
```diff
@@ -0,0 +1,68 @@
...
+	public abstract string performance_degraded { owned get; }
+
+	[DBus (name = "Profiles")]
+	public abstract GLib.HashTable<string, GLib.Variant>[] profiles { owned get; }
+
+	[DBus (name = "Actions")]
+	public abstract string[] actions { owned get; }
+
+	[DBus (name = "ActiveProfileHolds")]
+	public abstract GLib.HashTable<string, GLib.Variant>[] active_profile_holds { owned get; }
```
EbonJaeger, line 42: I don't think you need to specify the DBus names; Vala or DBus or whatever should convert them automagically.

src/panel/applets/status/PowerProfiles.vala
```diff
@@ -0,0 +1,68 @@
...
+
+	void on_proxy_get(Object? o, AsyncResult? res) {
+		try {
+			proxy = Bus.get_proxy.end(res);
+			proxy_connected();
+		} catch (Error e) {
+			warning("unable to connect to net.hadess.PowerProfiles- power-profiles-daemon is probably not installed. %s", e.message);
+		}
+	}
+}
```
EbonJaeger, line 68: I'm not sure this little class is necessary. I would suggest moving the `Bus.get_proxy` call to near the end of the constructor of PowerIndicator, and moving the `on_proxy_get` function to that class. In that function, you can then connect your signals and do the initial state setting.

EbonJaeger (changes requested):
Good start! My main concern is just removing the helper class that isn't really needed, since it doesn't do anything that can't just be done in the PowerIndicator class.

src/panel/applets/status/PowerIndicator.vala
justchen1369: ^ update: yep, it renders regardless of the existence of the bus for whatever reason

src/panel/applets/status/PowerIndicator.vala
```diff
@@ -238,6 +331,80 @@ public class PowerIndicator : Gtk.Bin {
...
+
+		SignalHandler.unblock(radio_power_save, radio_power_save_id);
+		SignalHandler.unblock(radio_power_balanced, radio_power_balanced_id);
+		SignalHandler.unblock(radio_power_performance, radio_power_performance_id);
+	}
+
+	void destroy_power_options() {
+		box.remove(power_profiles_box);
+		power_profiles_box = null;
+	}
```
EbonJaeger, line 406:
```suggestion
	void destroy_power_options() {
		power_profiles_box.destroy();
	}
```

EbonJaeger (changes requested):
Definitely closer! I think this can be simplified further, though. The `PowerProfilesProxy` object isn't really needed still. It's a proxy for a proxy, so just cut out the middleman. `PowerIndicator` can watch the Bus name, and then create and add the widget when it's acquired, and destroy it when it's lost.

If the profiles box widget is turned into its own widget class, it can then take in the resulting DBus interface and manage its own state, so it doesn't have to be in the main `PowerIndicator` object.

src/panel/applets/status/PowerIndicator.vala
```diff
@@ -146,11 +146,130 @@ public class BatteryIcon : Gtk.Box {
 	}
 }
 
+const string POWER_PROFILES_DBUS_NAME = "net.hadess.PowerProfiles";
+const string POWER_PROFILES_DBUS_OBJECT_NAME = "/net/hadess/PowerProfiles";
+
+namespace net.hadess {
+	[DBus (name = "net.hadess.PowerProfiles")]
+	public interface PowerProfilesDBus : Object {
+		[DBus (name = "Profiles")]
```
EbonJaeger, line 155: Once again, not necessary

src/panel/applets/status/PowerIndicator.vala
```diff
@@ -146,11 +146,130 @@ public class BatteryIcon : Gtk.Box {
...
+		public abstract HashTable<string, Variant>[] profiles { owned get; }
+
+		[DBus (name = "ActiveProfile")]
```
EbonJaeger, line 158: Same as above

src/panel/applets/status/PowerIndicator.vala
```diff
@@ -146,11 +146,130 @@ public class BatteryIcon : Gtk.Box {
...
+			return;
+		}
+
+		// initialize state
+		on_active_profile_changed(profiles_proxy.active_profile);
+
+		DBusProxy? proxy = profiles_proxy as DBusProxy;
+		proxy.g_properties_changed.connect(() => {
+			on_active_profile_changed(profiles_proxy.active_profile);
+		});
```
EbonJaeger, line 202:
```suggestion
		((DBusProxy) proxy).g_properties_changed.connect(() => {
			on_active_profile_changed(profiles_proxy.active_profile);
		});
```

src/panel/applets/status/PowerIndicator.vala
```diff
@@ -146,11 +146,130 @@ public class BatteryIcon : Gtk.Box {
...
+		header.set_halign(Gtk.Align.CENTER);
+		pack_start(header);
+
+		var power_profiles_radio_box = new Gtk.Box(Gtk.Orientation.HORIZONTAL, 1);
+
+		if (profiles.contains("power-saver")) {
+			radio_power_save = new Gtk.RadioButton.with_label_from_widget(null, _("Power Saver"));
+			radio_power_save_id = radio_power_save.toggled.connect(() => {
+				profiles_proxy.active_profile = "power-saver";
+			});
```
EbonJaeger, line 218: `toggled` is triggered both when the radio button is activated *and* deactivated, so be sure to check the current button state:

```suggestion
			radio_power_save_id = radio_power_save.toggled.connect(() => {
				if (radio_power_save.is_active()) {
					profiles_proxy.active_profile = "power-saver";
				}
			});
```

src/panel/applets/status/PowerIndicator.vala
```diff
@@ -146,11 +146,130 @@ public class BatteryIcon : Gtk.Box {
...
+			power_profiles_radio_box.pack_start(radio_power_save, false, false, 1);
+		}
+
+		if (profiles.contains("balanced")) {
+			radio_power_balanced = new Gtk.RadioButton.with_label_from_widget(radio_power_save, _("Balanced"));
+			radio_power_balanced_id = radio_power_balanced.toggled.connect(() => {
+				profiles_proxy.active_profile = "balanced";
+			});
```
EbonJaeger, line 226:
```suggestion
			radio_power_balanced_id = radio_power_balanced.toggled.connect(() => {
				if (radio_power_balanced.is_active()) {
					profiles_proxy.active_profile = "balanced";
				}
			});
```

src/panel/applets/status/PowerIndicator.vala
```diff
@@ -146,11 +146,130 @@ public class BatteryIcon : Gtk.Box {
...
+			power_profiles_radio_box.pack_start(radio_power_balanced, false, false, 1);
+		}
+
+		if (profiles.contains("performance")) {
+			radio_power_performance = new Gtk.RadioButton.with_label_from_widget(radio_power_save, _("Performance"));
+			radio_power_performance_id = radio_power_performance.toggled.connect(() => {
+				profiles_proxy.active_profile = "performance";
+			});
```
EbonJaeger, line 234:
```suggestion
			radio_power_performance_id = radio_power_performance.toggled.connect(() => {
				if (radio_power_performance.is_active()) {
					profiles_proxy.active_profile = "performance";
				}
			});
```

src/panel/applets/status/PowerIndicator.vala
```diff
@@ -196,12 +315,46 @@ public class PowerIndicator : Gtk.Bin {
...
 		toggle_show();
 	}
 
+	void has_power_profiles() {
+		if (profiles_proxy == null) {
+			Bus.get_proxy.begin<net.hadess.PowerProfilesDBus>(BusType.SYSTEM, POWER_PROFILES_DBUS_NAME, POWER_PROFILES_DBUS_OBJECT_NAME, 0, null, on_proxy_get);
+		} else {
+			// already have a proxy object
+			create_power_profiles_options();
+		}
```
EbonJaeger, line 332: I'm not sure that this will work if the connection was lost; the proxy object might be invalid. That's why I was thinking that the widget would store the proxy object so that this class doesn't have to.

src/panel/applets/status/PowerIndicator.vala
```diff
@@ -196,12 +315,46 @@ public class PowerIndicator : Gtk.Bin {
...
+		try {
+			profiles_proxy = Bus.get_proxy.end(res);
+
+			if (profiles_proxy.active_profile != null) {
+				create_power_profiles_options();
+			}
+
+		} catch (Error e) {
+			warning("unable to connect to net.hadess.PowerProfiles: %s", e.message);
+			return;
```
EbonJaeger, line 348: Unnecessary return at end of function/flow control

src/panel/applets/status/PowerIndicator.vala
```diff
@@ -146,11 +146,130 @@ public class BatteryIcon : Gtk.Box {
...
+		orientation = Gtk.Orientation.VERTICAL;
+		
+		var profiles = new GenericSet<string>(str_hash, str_equal);
+
+		foreach (HashTable<string, Variant> profile in profiles_proxy.profiles) {
+			var profile_value = profile.get("Profile");
+
+			if (!profile_value.is_of_type(VariantType.STRING)) {
+				warning("unexpected type for power profile: is not of VariantType.STRING");
+				return;
```
EbonJaeger, line 184: Skip to the next item instead of exiting:

```suggestion
				continue;
```

src/panel/applets/status/PowerIndicator.vala
```diff
@@ -196,12 +315,46 @@ public class PowerIndicator : Gtk.Bin {
...
 
+	void has_power_profiles() {
+		if (profiles_proxy == null) {
+			Bus.get_proxy.begin<net.hadess.PowerProfilesDBus>(BusType.SYSTEM, POWER_PROFILES_DBUS_NAME, POWER_PROFILES_DBUS_OBJECT_NAME, 0, null, on_proxy_get);
+		} else {
+			// already have a proxy object
+			create_power_profiles_options();
+		}
+	}
+	void lost_power_profiles() {
```
EbonJaeger, line 334: Missing empty line between functions

EbonJaeger (changes requested):
Now we're talkin'! Getting down to much more minor changes now. :)

src/panel/applets/status/PowerIndicator.vala
justchen1369: tested- it does actually work

src/panel/applets/status/PowerIndicator.vala
justchen1369: the handlers for all three are blocked and then unblocked in `on_active_profile_changed()`, so that shouldn't be necessary

src/panel/applets/status/PowerIndicator.vala
justchen1369: ditto

EbonJaeger (approved):
Overall, good work! Thank you for taking the time to address the various concerns. I'll approve this now and kick it over to Josh so he can look it over and give the final approval (or request more changes).

src/panel/applets/status/PowerIndicator.vala
```diff
@@ -146,11 +146,130 @@ public class BatteryIcon : Gtk.Box {
...
+	public PowerProfilesSelector(PowerProfilesDBus profiles_proxy) {
+		orientation = Gtk.Orientation.VERTICAL;
+		
+		var profiles = new GenericSet<string>(str_hash, str_equal);
+
+		foreach (HashTable<string, Variant> profile in profiles_proxy.profiles) {
+			var profile_value = profile.get("Profile");
+
+			if (!profile_value.is_of_type(VariantType.STRING)) {
+				warning("unexpected type for power profile: is not of VariantType.STRING");
```
JoshStrobl, line 178: Not likely to get an out-of-spec variant from PPD so can probably drop this.

```suggestion
```

src/panel/applets/status/PowerIndicator.vala
```diff
@@ -146,11 +146,130 @@ public class BatteryIcon : Gtk.Box {
...
+
+		if (profiles.contains("performance")) {
+			radio_power_performance = new Gtk.RadioButton.with_label_from_widget(radio_power_save, _("Performance"));
+			radio_power_performance_id = radio_power_performance.toggled.connect(() => {
+				if (radio_power_performance.get_active()) {
+					profiles_proxy.active_profile = "performance";
+				}
+			});
+			power_profiles_radio_box.pack_start(radio_power_performance, false, false, 1);
+		}
```
JoshStrobl, line 236: I think a fair bit of this code is duplicated and we can instead create class that takes the power proxy, type, etc. and creates a RadioButton that when toggled, fires off the type to the provided proxy.

src/panel/applets/status/PowerIndicator.vala
```diff
@@ -196,12 +315,46 @@ public class PowerIndicator : Gtk.Bin {
...
+	}
+
+	void on_proxy_get(Object? o, AsyncResult? res) {
+		try {
+			profiles_proxy = Bus.get_proxy.end(res);
+
+			if (profiles_proxy.active_profile != null) {
+				create_power_profiles_options();
+			}
+
```
JoshStrobl, line 346: Unnecessary newline and can probably do:

```suggestion
			if (profiles_proxy.active_profile != null) create_power_profiles_options();

```

src/panel/applets/status/PowerIndicator.vala
```diff
@@ -196,12 +315,46 @@ public class PowerIndicator : Gtk.Bin {
...
 		toggle_show();
 	}
 
+	void has_power_profiles() {
+		if (profiles_proxy == null) {
+			Bus.get_proxy.begin<PowerProfilesDBus>(BusType.SYSTEM, POWER_PROFILES_DBUS_NAME, POWER_PROFILES_DBUS_OBJECT_NAME, 0, null, on_proxy_get);
+		} else {
+			// already have a proxy object
+			create_power_profiles_options();
+		}
```
JoshStrobl, line 332:
```suggestion
		if (profiles_proxy != null) return create_power_profiles_options();
		Bus.get_proxy.begin<PowerProfilesDBus>(BusType.SYSTEM, POWER_PROFILES_DBUS_NAME, POWER_PROFILES_DBUS_OBJECT_NAME, 0, null, on_proxy_get);
```

src/panel/applets/status/PowerIndicator.vala
```diff
@@ -324,12 +311,9 @@ public class PowerIndicator : Gtk.Bin {
 	}
 
 	void has_power_profiles() {
-		if (profiles_proxy == null) {
-			Bus.get_proxy.begin<PowerProfilesDBus>(BusType.SYSTEM, POWER_PROFILES_DBUS_NAME, POWER_PROFILES_DBUS_OBJECT_NAME, 0, null, on_proxy_get);
-		} else {
-			// already have a proxy object
-			create_power_profiles_options();
-		}
+		if (profiles_proxy != null) create_power_profiles_options();
```
JoshStrobl, line 314: You're missing a return here. With the code now, it will always call to get the proxy even when it is unnecessary

JoshStrobl (changes requested):
LGTM otherwise :)

PR #354, opened 2023-04-13
Add clangd cache to gitignore

serebit:
## Description

clangd can be used with Visual Studio Code, and we definitely shouldn't be committing its cache.

### Submitter Checklist


JoshStrobl (approved):
LGTM, thanks!

Issue #28, opened 2022-01-06
Evaluate potential third-party widgets to merge into mainline


JoshStrobl:
This is a proposal to evaluate third-party widgets from folks like Ubuntu Budgie, reducing duplicate applets by consolidating functionality (such as DropBy into Places), and enabling more of these applets to be enjoyed by Budgie users without supplemental work.

Other examples:

- @EbonJaeger's budgie-track-applet and Ubuntu Budgie's trash applet (you folks must really like trash?)
- Take a Break (perfect for Caffeine)
- Quick Note
- Keyboard Auto Switch into our input switcher
- FuzzyClock may be able to be deprecated given we support custom formats now too
- VisualSpace (maybe as a replacement for the Wnck-based Workspace applet)?

Thoughts @fossfreedom, @serebit, @Jacob-Vlijm, @fossfreedom, etc?

serebit:
Quick Note, yes. Trash, absolutely. No opinion on others.

EbonJaeger:
I'd have no issue mainlining my Budgie Trash Applet, if so desired. If we want a trash applet in Budgie, we'd just have to decide on which one, assuming the Ubuntu Budgie folks are also okay with it.

Another consideration is that mine is in C and I believe theirs is in Vala; I'm not exactly familiar with theirs.

It's a trashy job, but someone's gotta do it. (You know I couldn't resist making a trash-related pun xD)

fossfreedom:
I haven't got a firm view yet on the suggestions here.

My general view is that yes we should encourage unique and well defined and designed applets that have a wide appeal within a thriving ecosystem.

Should this be under one banner? Not sure. Most applet developers we have seen have developed stuff because they had an itch to scratch rather than thinking through the various issues such as translation support, well documented code, consistent design etc.

Applet devs are a unique bunch! I would not want them to face perceived barriers where they could not directly make changes to what they may deem as "theirs" since they wrote 99% of it.

Maybe splitting applets into an applet repo with donaters/devs having full control to keep developing ... and maybe helping each other out?
 Don't really know the answer.

JoshStrobl:
@fossfreedom I think that is some good insight and I do appreciate it. I fully recognize that some people might not even want to mainline an applet, just as how I entirely recognize that may be valuable to have some remain "downstream" to continue to provide even more value-add for that operating system. Both of those reasons are completely acceptable.

[...]

PR #355, opened 2023-04-19
Upstreaming budgie-trash-applet

EbonJaeger:
## Description
This moves my Budgie Trash Applet into the Budgie Desktop source tree.

As a part of this, the applet has been redesigned and rearchitected to better fit within Budgie Desktop:

* The header now looks similar to the in-progress Bluetooth indicator and network applet
* Buttons to restore items or empty the trash bin have been added under the header
* The concept of separate trash bins has been removed; there is just one global trash bin
* The timestamp when a file was trashed is now shown under the file's name
* Settings button has been moved to the header

In addition to the visual changes, some of the UX have been changed. The restore button no longer just restores all items in the trash; it only restores selected items. This makes it easier to restore just one file, or to quickly restore several files at once.

Also, applet settings have been changed to be per-instance, so different instances of the applet on your panels will all have their own settings, just in case you actually want that.

Lastly, upstreaming the applet more easily allows the applet to be translated into all the wonderful languages spoken by our community. Thank you, translators, for your valuable contributions! :)

Ref #28 

### Submitter Checklist


EbonJaeger:
Speaking of translations, I entirely forgot to write that code before making the PR. Derp.

serebit:
CI failed—you're linking against a system header for popover.h instead of the one in the source tree.

serebit (approved):
LGTM, thanks!

JoshStrobl (approved):
LGTM :rocket:

PR #363, opened 2023-04-24
fix: check of zenity on OSes with older Meson

JoshStrobl:
This PR fixes zenity detection on OSes which do not have a new enough meson to have its `version()` check.


JoshStrobl:
Merging with permission from @fossfreedom

PR #370, opened 2023-04-29
Ensure that only valid nodes are added to tray item menus

serebit:
## Description

This fixes segfaults caused by some apps that have tray icons. I believe that it happens due to bad implementations of the dbusmenu spec on the app side, which then cause the applet's internal record of the menu layout to be filled with invalid data. Solution is to completely ignore any items that don't have the correct type for the layout id (it should be int32), and also avoid showing menus without any items in them. This is reproducible with JetBrains Toolbox.

Needs review from @JoshStrobl, who I believe was seeing this happen with a different app.

### Submitter Checklist

### DIFF
--- a/src/panel/applets/tray/DBusMenu.vala
+++ b/src/panel/applets/tray/DBusMenu.vala
@@ -66,8 +66,15 @@ public class DBusMenu : Object {
 		all_nodes.get(0).submenu.show_all();
 	}
 
-	private DBusMenuNode parse_layout(Variant layout) {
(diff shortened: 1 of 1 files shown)


JoshStrobl (approved):
LGTM! Unfortunately haven't been able to reproduce it with the same autostarting apps that triggered it the first time, so I get the feeling a race condition of some sort is affecting it, but the PR looks good so gonna land it :) Thanks! :gem:

PR #381, opened 2023-05-21
Remove style class changes in power dialog

EbonJaeger:
## Description

This fixes the case where themes that don't style the dialog end up having a transparent background, making it very hard to see and use the dialog.

One thing to be aware of is that the previous way was done to combat focus issues when presenting the dialog under certain conditions. So far, though, I have been unable to get the focus to mess up with the new changes; the dialog is always focused correctly.

### Submitter Checklist


JoshStrobl (approved):
LGTM :rocket:

PR #387, opened 2023-05-29
Add support for Magpie

fossfreedom:
## Description
Add build support for magpie

WIP - requires "converttomagpie" branch of magpie repo

### Submitter Checklist


serebit:
Why not generate the vapi on the Magpie side instead?

fossfreedom:
@serebit possibly.  Not sure if it would be just a little confusing from a "what we have done before" POV.

JoshStrobl (approved):

PR #388, opened 2023-05-30
Change the sorting of some applications and categories

EbonJaeger:
### Rationale

These changes are based on user feedback and just having used the current category sorting for an extended period of time. Applications that previously were put into the Other category fit better in other existing categories. Some of the 
application exclusions just no longer make sense.

#### Utilities Category

This category has been removed entirely. It had a weird overlap with the System Tools category, and it wasn't clear what the difference was between Utilities and Accessories, leading to confusion. To address this, items that used to be in the Utilities category will now be included in the System Tools category.

### Submitter Checklist


JoshStrobl (approved):
LGTM but let's monitor to see if anything manages to sneak into a weird category that might warrant migration.

PR #391, opened 2023-06-03
Provide more information in privilege escalation dialogs

### LINKED ISSUE #390: Provide more information in privilege escalation dialogs
The_Machine_101
Mar 26
I have observed after updating application that this authentication is required and presented via pop-up. There are no details provided as to what is requesting the authentication.

I would like to see a details tab on the pop-up detailing what application or process is requiring this (or any other) authentication as it would be beneficial to the security of the OS for all users. This detail is available on other platforms from the pop-up (dialogue box).

I have been referred here by fossfreedom and I attach the following:

fossfreedom
Ubuntu Budgie Team Member
Mar 26
Can you please drop a screenshot with what you are seeing?

The_Machine_101

4h
This is a screenshot of the pop-up:
image

This is what I would like to see (two examples):
image
image

EbonJaeger:
## Description

This will show the action description and action ID in the Polkit authentication dialog when graphical privilege escalation is requested.

The Polkit Vapi files that we previously used have also been nuked, since they are a part of the base Vala installation.

Fixes #390 

### Submitter Checklist


src/dialogs/polkit/dialog.ui
```diff
@@ -147,15 +285,18 @@
...
           <object class="GtkButtonBox" id="buttonbox1">
             <property name="visible">True</property>
-            <property name="can_focus">False</property>
-            <property name="layout_style">expand</property>
+            <property name="can-focus">False</property>
+            <property name="layout-style">expand</property>
             <child>
-              <object class="GtkButton" id="button_cancel">
-                <property name="label" translatable="yes">Cancel</property>
+              <object class="GtkButton" id="button_auth">
```
JoshStrobl, line 291: Can we flip the buttons back around plz?

EbonJaeger: I don't even remember changing that in the first place :o

JoshStrobl (approved):
LGTM!

PR #392, opened 2023-06-03
Drop all Vapi files that are already included with Vala

EbonJaeger:
## Description

Just what it says on the tin.

Note: The polkit Vapi files aren't deleted in this PR because they'll be removed with #391.

### Submitter Checklist


JoshStrobl (approved):
Hell yes.

PR #393, opened 2023-06-13
Hide Kvantum and Murrine themes from budgie-desktop-settings

BlindRepublic:
## Description
This commit adds the prefixes "Kv" and "Murrina" to the list of blacklisted themes since they aren't related to GTK3 but are still listed in Budgie Desktop Settings.

### Submitter Checklist


serebit:
What do they look like when enabled?

BlindRepublic:
> What do they look like when enabled?

They look like stock adwaita which is already hidden

JoshStrobl (approved):
LGTM, thanks!

PR #398, opened 2023-07-04
Change default systray spacing from 4 to 2

serebit:
## Description

This matches the default panel spacing and makes the panel icons look a lot more uniform now that Status Notifier is in.

### Submitter Checklist


JoshStrobl (approved):
LGTM, thanks!

PR #400, opened 2023-07-05
Use ngettext for plural strings

khagaroth:
Use ngettext plural support to allow proper translations.

### DIFF
--- a/src/daemon/notifications/dbus.vala
+++ b/src/daemon/notifications/dbus.vala
@@ -554,11 +554,7 @@
 			var summary = _("Unread Notifications");
 
 			string? body = null;
-			if (paused_notifications == 1) {
-				body = _("You received 1 notification while an application was fullscreened.");
-			} else {
-				body = _("You received %d notifications while an application was fullscreened.".printf(this.paused_notifications));
-			}
+			body = ngettext("You received %d notification while an application was fullscreened.", "You received %d notifications while an application was fullscreened.", this.paused_notifications).printf(this.paused_notifications);
 
 			var icon = "dialog-information";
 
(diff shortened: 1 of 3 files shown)


src/daemon/settings.vala
```diff
@@ -341,11 +341,7 @@ namespace Budgie {
...
 
 				if (enabled && (time > 0)) {
-					var duration = _("Will turn off in a minute");
-
-					if (time > 1) { // use plural
-						duration = _("Will turn off in %d minutes").printf(time);
-					}
+					var duration = ngettext("Will turn off in a minute", "Will turn off in %d minutes", time).printf(time);
 
 					body = duration;
```
JoshStrobl, line 346: Can just have body = ngettext(...)

src/daemon/notifications/dbus.vala
```diff
@@ -569,11 +569,7 @@
 			var summary = _("Unread Notifications");
 
 			string? body = null;
-			if (paused_notifications == 1) {
-				body = _("You received 1 notification while an application was fullscreened.");
-			} else {
-				body = _("You received %d notifications while an application was fullscreened.".printf(this.paused_notifications));
-			}
+			body = ngettext("You received %d notification while an application was fullscreened.", "You received %d notifications while an application was fullscreened.", this.paused_notifications).printf(this.paused_notifications);
```
JoshStrobl, line 572: Can just have `string body = ngettext(...)`

JoshStrobl (changes requested):
Otherwise LGTM. I appreciate you taking the initiative to improve this! :heart:

src/daemon/notifications/dbus.vala
serebit: This line is also *really* long, and should probably be broken up into multiple lines, like so:

```vala
body = ngettext(
    "You received %d notification while an application was fullscreened.",
    "You received %d notifications while an application was fullscreened.",
    this.paused_notifications
).printf(this.paused_notifications);
```

JoshStrobl:
Went ahead and fixed my own PR items. WIll merge when CI is done.

PR #404, opened 2023-07-09
Search for plugins in both /usr and /usr/local

serebit:
## Description

This was trickier to implement *well* than I'd like, but I think I stumbled onto a reasonable solution.

If the set prefix is /usr, a "secondary" plugin search directory is added that uses /usr/local. If the set prefix is /usr/local, that secondary plugin search directory uses /usr instead. Otherwise, only the set prefix is used for plugins. 

### Submitter Checklist


JoshStrobl (approved):
LGTM, thanks!

PR #406, opened 2023-07-09
Prevent notifs and scrshot from locking up when sink is locked

serebit:
## Description

The two cases are handled differently, as screenshots use gstreamer and notifs use canberra.

The screenshot sound now has a timeout of 2 seconds, after which it will cancel the attempt to play the sound and just move forward with the screenshot. This may cut off longer screenshot sounds, but it doesn't cut off the default.

The notif sounds are now launched as threads, so they don't block the main thread when queued up. This means that all buffered notif sounds will play upon the default sink being unlocked, but this is how all other sound players work, so it's not a big deal.

### Submitter Checklist

### DIFF
--- a/src/daemon/notifications/dbus.vala
+++ b/src/daemon/notifications/dbus.vala
@@ -471,28 +471,13 @@
 				sound_name = notification.hints.get("sound-name").get_string();
 			}
 
-			play_sound(notification, sound_name);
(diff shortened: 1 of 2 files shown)


JoshStrobl (approved):
LGTM, thanks!

PR #407, opened 2023-07-10
Use is_empty() instead of length checks on collections

EbonJaeger:
## Description
This is finally possible now that we no longer use our own generated Vapi files for GLib.

### Submitter Checklist


EbonJaeger:
Also, sorry in advance to any rebasing this causes.

JoshStrobl (approved):
LGTM, thanks!

PR #410, opened 2023-07-12
trash-applet: Try to use GFile wherever possible

### LINKED ISSUE #409: [Bug] Trash applet unable to act on items with a "%" in the name
### Description

After building the latest budgie desktop from source, Trash applet has difficulties dealing with files that have a "%" in the name. While they show up in the popover, the actions return error messages when attempting to handle these files.

### Budgie version

git f59e6cf622fc5d2befbcccfd75e2d1fc3a563fa2

### Operating System

Ubuntu 23.10

1. Create a file with a "%" in the file name (i.e "10%")
2. Drag and drop the file on the trash applet (Error message says not supported)
3. Move the file to the trash using a file manager - The file shows up in the trash applet
4. Select the file and attempt to delete it

### Actual result

Error message saying "Unable to delete (null): Operation not supported"

ss

### Expected result

The file to be restored or deleted

### Additional information

_No response_

EbonJaeger:
## Description

As long as we have valid GFile objects, file operations won't fail. The trick is getting correct GFiles when initially scanning the trash directory, because GLib in its infinite wisdom makes GFileEnumerator return a list of GFileInfo, not GFile. And file name/path URI encoding is a mess.

Tested with:
- Files with % in the name, both on local and removable drives
- Files without % in the name, both on local and removable drives
- Files with spaces in the name, both on local and removable drives
- Files without spaces in the name, both on local and removable drives
- Files only consisting of letters, both on local and removable drives

Tested by:
- Drag-n-dropping file to trash applet button
- Trashing file in file manager
- Permanently deleting file via applet
- Restoring file via applet

Fixes #409

### Submitter Checklist

### DIFF
--- a/src/panel/applets/trash/trash_applet.c
+++ b/src/panel/applets/trash/trash_applet.c
@@ -215,18 +215,18 @@ static void drag_data_received(
 	g_return_if_fail(info == 0);
 
 	g_autofree gchar *uri = g_strdup((gchar *) gtk_selection_data_get_data(data));
-	g_autofree gchar *unescaped = NULL;
 	g_autoptr(GFile) file = NULL;
 	g_autoptr(GError) err = NULL;
 
 	if (g_str_has_prefix(uri, "file://")) {
-		unescaped = g_uri_unescape_string(uri, NULL);
-		g_strstrip(unescaped); // Make sure there's nothing silly like a trailing newline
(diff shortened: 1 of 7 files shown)


JoshStrobl (approved):
LGTM, thanks!

Issue #372, opened 2023-05-01
[Bug] appindicators is broken due to statusnotifier starting before appindicator initialisation


fossfreedom:
### Description

Recording this to be looked at as to the best way to handle.

Reported by the UB team with the daily budgie desktop PPA

Appindicators via the applet no longer display application indicators despite the new notifier system tray not being on the panel.

Status notifier is starting before appindicators and taking control of the same dbus that appindicators require

Recording this as an issue to be worked through by the UB team with input from buddies.

Current thoughts in no particular order

1. Disable Status notifier as part of the build and add appindicators as a package dependency for Debian and Ubuntu
2. Remove appindicator support entirely from Debian and Ubuntu. 
3. Add the ability somewhere to unhook/hook Status notifier from starting/to start  as part of budgie initialisation. Needs some thought how best to express this via GUI.
4. Rework the appindicator applet to use Status notifier hooks. Doubtful if viable due to the dependency chain on the appindicator server but worth a look.
5. Something else

### Budgie version

Main

### Operating System

UB 23.10 plus git main

### Steps to reproduce the issue

1. Add appindicator applet to panel
2. Remove status notifier system tray
3. Logout and login

### Actual result

4. Note that no appindicators are visible despite running

### Expected result

Allow either/or status notifier and appindicators on panel

serebit:
My opinion:
- Option 1 is a complete no-go. Changing the Budgie configuration is fine, but making the Budgie API surface dependent on the operating system is not.
- Option 2 would be my recommendation—AppIndicator items work with my StatusNotifier implementation, so not much reason to keep a dedicated AppIndicator dep and applet around.
- Option 3 seems like a bad idea for similar reasons to option 1.
- Option 4 would be my second choice. If you *really* want AppIndicator, this would be the way to do it without changing the Budgie API surface.

fossfreedom:
closing - managing this as an in-distro patch.

PR #414, opened 2023-07-17
Allow AppIndicator to run if System Tray is not present

samlane-ma:
## Description
Allow Ubuntu Budgie's default AppIndicator applet to work, if System Tray is not on the panel

In regards to #372, this is a possible solution to allow Ubuntu Budgie's appindicator applet to properly run. It works by checking the panels to see if System Tray or AppIndicator is present. In a case where AppIndicator is present on a panel, and System Tray is not, it prevents the initialization of Status Notifier. In order to avoid any conflicts with the default Budgie Desktop experience, if both applets are present, or neither applet is present, Status Notifier will still initialize as usual.

### Submitter Checklist

### DIFF
--- a/src/daemon/manager.vala
+++ b/src/daemon/manager.vala
@@ -32,7 +32,7 @@ namespace Budgie {
 		*/
 		public ServiceManager(bool replace) {
 			theme_manager = new Budgie.ThemeManager();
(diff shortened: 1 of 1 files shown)


serebit:
I appreciate the attempt to resolve this conflict, but this isn't a reasonable solution. Closing.

PR #418, opened 2023-08-02
keyboard-layout: Don't try to initialize IBus

EbonJaeger:
## Description
IBus already gets initialized in wm, and calling that function here causes severe breakage, leading to budgie-panel being repeatedly killed for OOM.

### Submitter Checklist


JoshStrobl (approved):
LGTM, thanks!

PR #420, opened 2023-08-09
Update Copyright Years

acook:
Updates the readme and version string.


JoshStrobl (changes requested):
The years can just be removed. See 0eea3fd

JoshStrobl (approved):
Thanks!

PR #423, opened 2023-08-14
rename misleading header

justchen1369:


JoshStrobl:
No. I intentionally changed this since it affects performance and would be more understandable by users.

PR #443, opened 2023-09-03
Add xdg-desktop-portal config

### LINKED ISSUE #436: [Request] ship an appropriate portals.conf for xdg-desktop-portal v1.17
### Budgie version

10.8

### Use-cases

Define a suitable portal for budgie

### Proposal

Debian and probably other distros are going to request desktop environments to ship a suitable portals.conf file so that it chooses the appropriate xdg-desktop-portal. So for us it probably should be the gtk portal rather than the gnome portal

More information below released by debian and in the link it gives a suggestion for the budgie portal.conf file

### References

Package: cinnamon-common
Severity: normal
Tags: trixie sid
User: [xdg-desktop-portal@packages.debian.org](mailto:xdg-desktop-portal@packages.debian.org)
Usertags: portals.conf

xdg-desktop-portal 1.17.x introduces a new way to select which portals will
be used for which desktop environments, modelled on mimeapps.list:

- each desktop environment should provide a file like
  /usr/share/xdg-desktop-portal/x-cinnamon-portals.conf

- the filename is ${DESKTOP}-portals.conf where ${DESKTOP} is the desktop
  environment's entry in $XDG_CURRENT_DESKTOP (the same as the DesktopNames
  from /usr/share/{x,wayland-}sessions/*.desktop), folded to lower case

- sysadmins and users can override this via files named portals.conf or
  ${DESKTOP}-portals.conf in various locations like /etc/xdg-desktop-portal
  and ~/.config/xdg-desktop-portal

Please see portals.conf(5) or its source code
https://github.com/flatpak/xdg-desktop-portal/blob/main/doc/portals-conf.rst
for full details.

As a backwards-compatibility mechanism, x-d-p will fall back to trying
to guess the most appropriate portals from the portals' UseIn= fields,
but it will log warnings when it does that, and anyway Debian doesn't
currently ship any portal backends that are flagged as suitable for Cinnamon
(although I see there's an ITP open for x-d-p-xapp, #1038946). Please
add an x-cinnamon-portals.conf to tell x-d-p more explicitly what backends
Cinnamon is meant to be using by default.

For example, if the intention is to try to use the -xapp backend, falling
back to -gtk if -xapp isn't installed or doesn't know how to do something,
the way to write that would be:

[preferred]
default=xapp;gtk;

The desktop environment (either cinnamon-common or some larger metapackage)
should probably also have a Recommends, or at least a Suggests, on whatever
portals would be most appropriate for it.

Thanks,
    smcv

--
This is part of a mass bug filing:
https://lists.debian.org/debian-devel/2023/08/msg00311.html

serebit:
## Description

Defaults to xdg-desktop-portal-gtk for all portals. Resolves #436. xdg-desktop-portal v1.17.0 isn't stable, so this PR is untested.

### Submitter Checklist

### DIFF
--- a/data/budgie-portals.conf
+++ b/data/budgie-portals.conf
@@ -1,0 +1,3 @@
+[preferred]
+# Use xdg-desktop-portal-gtk for every portal interface
+default=gtk
--- a/data/meson.build
+++ b/data/meson.build
@@ -15,3 +15,8 @@ install_data(
     version_file,
     install_dir: join_paths(datadir, 'budgie')
 )
+
+install_data(
+    'budgie-portals.conf',
+    install_dir: join_paths(datadir, 'xdg-desktop-portal')
+)


JoshStrobl (approved):
LGTM, thanks!

PR #444, opened 2023-09-04
Make dark theme switch apply to color scheme pref

serebit:
## Description

Adds a few LoC into the daemon that synchronize the dark theme dconf setting with GNOME's freedesktop color scheme dconf setting.

### Submitter Checklist


JoshStrobl (approved):
LGTM, thanks!

PR #452, opened 2023-09-09
Simplify Raven notification group

### LINKED ISSUE #450: [Bug] Raven notifications widget sometimes shows broken icon placeholder
### Description

Sometimes when displaying notifications, Raven will show the broken image placeholder instead of the actual icon the notification used.

[example screenshot](https://postimg.cc/dkS6djpQ)

My example below to reproduce uses notify-send, but I've also seen it in a few apps, including streamlink-twitch-gui. (Go to settings, notifications, and click on 'test notification')

I should note that the notification itself shows the correct icon, the broken icon image is only in Raven's widget for missed notifications.  It also does not appear to be a directory permissions issue.  I think that Raven may simply not be able to show PNG icons correctly there, SVG icons placed in .local/share/icons or /usr/share/icons seem to work correctly but PNGs in those directories do not.

### Budgie version

10.8

### Operating System

Solus 4.4

Works correctly:
`notify-send test`
`notify-send --icon=volume-low test`
`notify-send --icon=<SVG icon in ~/.local/share/icons> test`

Works incorrectly:
`notify-send --icon=<PNG icon in ~/.local/share/icons or /usr/share/icons> test`

### Actual result

[example screenshot](https://postimg.cc/dkS6djpQ)

### Expected result

correctly shows notification icon

### Additional information

none

EbonJaeger:
## Description
This does a couple of things. First, when constructing a notification group, just get the image directly from our Notification object. We did all the work to get the correct image, let's actually use it.

Secondly, remove our HashTable mapping application name to notification group. We don't need it; anything it was used for we can get from the ListBox. This avoids having to keep track of groups in multiple data structures (the table and the list box). This makes the code easier to maintain, and slightly more memory efficient.

The NotificationGroup widget has been changed to a `Gtk.ListBoxRow` to make it easier to find and remove groups from the view.

Lastly, switch to `ngettext` to build the text for the notification view header instead of an if/elseif chain.

Fixes #450 

### Submitter Checklist

### DIFF
--- a/src/daemon/notifications/popup.vala
+++ b/src/daemon/notifications/popup.vala
@@ -301,7 +301,8 @@ namespace Budgie.Notifications {
 				margin_bottom = 5,
 				margin_right = 16,
 				halign = Gtk.Align.START,
-				hexpand = true
+				hexpand = true,
+				use_markup = true
 			};
 			title_label.get_style_context().add_class("notification-title");
 
--- a/src/lib/notification.vala
+++ b/src/lib/notification.vala
@@ -73,9 +73,15 @@
 			HashTable<string, Variant> hints,
 			uint expire_timeout
 		) {
+			var name = app_name;
+
+			if (("budgie" in name) && ("caffeine" in app_icon)) { // Caffeine Notification
+				name = _("Caffeine Mode");
+			}
+
 			Object (
 				id: id,
-				app_name: app_name,
+				app_name: name,
 				app_icon: app_icon,
 				summary: summary,
 				body: body,
--- a/src/lib/notification.vala
+++ b/src/lib/notification.vala
@@ -47,9 +47,9 @@
 
 		/* Notification information */
 		public string app_name { get; construct; }
+		public string notification_icon { get; construct; }
 		public HashTable<string, Variant> hints { get; construct; }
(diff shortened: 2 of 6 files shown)


JoshStrobl:
Testing with Discord and seems a bit busted. Incorrect title and notification group icons aren't that of the application, just of the first icon of the first notification you get.

image

serebit (changes requested):
Built-in theme needs a style update to coincide with this, the new notification items look off with the existing styling.

JoshStrobl:
@serebit Could you clarify what styling changes you are referring to?

- Notification Widget has not seen any changes with this PR. The notification widget clones attributes from the Notification itself.
- Notification Group, which has seen changes, is already referenced in our internal styling by its class name `raven-notifications-group`. The Raven Notifications Group Header is still aligned with group item headers, which was the purpose of margin and padding offsets, and the use of the inner box instead of the group itself being the box is still valid as we referenced the group without a parent selector (so it can be any child) and the inner list is still a valid selector
- Notification View is referenced by `raven-notifications-view` , saw no related changes from what I can tell.

It looks identical to me.

JoshStrobl (approved):
LGTM!

serebit (approved):

PR #455, opened 2023-09-11
Attempt to unbox primitives for DBusMenu nodes

### LINKED ISSUE #451: System tray menus are sometimes blank
### Description

Sometimes when opening the menu of a system tray indicator, the menu entries have no text. They're still there, and clicking them still works, just no text is displayed. So far I haven't found a reliable way to reproduce it. My initial thought is that it's related to items in the menu changing, but that's mostly just a guess at this point.

### Budgie version

10.8

### Operating System

Solus 4.4

1. Right click network-manager indicator
2. Disconnect from or connect to a network
3. Re-open indicator menu

### Actual result

Got it to happen once after disconnecting from the network:

image

### Expected result

image

### Additional information

_No response_

serebit:
## Description

No idea why, but sometimes properties are boxed inside variants and need unboxing before they can be used. When a property is updated, if the value is a variant type, attempt to unbox it first up to a limit of 8 times. Resolves #451.

### Submitter Checklist

### DIFF
--- a/src/panel/applets/tray/DBusMenuNode.vala
+++ b/src/panel/applets/tray/DBusMenuNode.vala
@@ -294,43 +294,43 @@ private class Properties {
 		switch (key) {
 			case "visible":
 				var old_value = visible;
-				visible = Properties.parse_bool(value, true);
+				visible = Properties.parse_bool(unbox(value), true);
 				return visible != old_value;
 			case "enabled":
 				var old_value = enabled;
-				enabled = Properties.parse_bool(value, true);
+				enabled = Properties.parse_bool(unbox(value), true);
 				return enabled != old_value;
 			case "label":
 				var old_value = label;
-				label = Properties.parse_string(value, "");
+				label = Properties.parse_string(unbox(value), "");
 				return label != old_value;
 			case "type":
 				var old_value = type;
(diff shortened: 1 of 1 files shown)


src/panel/applets/tray/DBusMenuNode.vala
```diff
@@ -378,4 +378,13 @@ private class Properties {
 
 		return ret;
 	}
+
+	private static Variant? unbox(Variant? variant) {
+		int levels = 0; // insurance policy
```
EbonJaeger, line 383: I'm curious, why 8?

EbonJaeger (commented):
Seems to fix the menu oddities I've observed, though I could only reproduce them only rarely.

src/panel/applets/tray/DBusMenuNode.vala
EbonJaeger: Oh, wrong line. Oh well, you know what I mean.

serebit: Didn't want it to be infinite, but also didn't want to only unbox once, just in case. 8 seemed like a decent cap.

EbonJaeger (approved):
LGTM! Seems to be working just as advertised.

PR #459, opened 2023-09-23
Implement a new Bluetooth indicator

### LINKED ISSUE #197: Implement a new Bluetooth item
### Proposal
Currently, we use gnome-bluetooth pretty directly, so we are beholden to what they do upstream. This recently became a slight pain point with the release of gnome-bluetooth 3, which is a whole new API.

It has been proposed to instead write a new implementation in Budgie using Bluez directly. This allows us to have full control over the entire implementation, further decoupling Budgie from GNOME.

The new implementation should have support for turning Bluetooth on and off, connecting and disconnecting paired devices, going into pairing mode, sending files over Bluetooth to a device, and display power levels of connected Bluetooth devices.

### References
#82

EbonJaeger:
## Description

The Bluetooth rewrite is finally here! GNOME Bluetooth is entirely dropped from Budgie Desktop; we now use Bluez and UPower directly, giving us full control over how Bluetooth is interacted with. You will now be able to connect to and disconnect from your Bluetooth devices from the panel without having to go to the Control Center. You will still have to visit the Control Center to pair new devices, however. With the direct usage of UPower, you will also see the battery power of devices that report it right in the applet popover.

This also includes our own implementation of Sendto in order to send and receive files over Bluetooth. Budgie Sendto is very heavily inspired by Elementary's own Bluetooth Sendto implementation. It runs in the background to listen for file transfer requests from other devices, showing a notification when one is received, enabling you to accept or reject the transfer. To send a file, you can open the Bluetooth indicator popover and hit the Send button to open a dialog to choose what device you want to send to.

Closes #197 

### Submitter Checklist

### DIFF
--- a/src/dialogs/sendto/Dialog/FileReceiver.vala
+++ b/src/dialogs/sendto/Dialog/FileReceiver.vala
@@ -10,6 +10,7 @@
  */
 
 public class FileReceiver : Gtk.Dialog {
+	public Bluetooth.Device device { get; set; }
 	public string session_path { get; set; }
 
 	public Bluetooth.Obex.Transfer transfer;
@@ -24,7 +25,7 @@ public class FileReceiver : Gtk.Dialog {
 
 	private Notification notification;
 
-	private string folder_path = "";
+	private string file_name = "";
 	private int start_time = 0;
 	private uint64 total_size = 0;
 
@@ -138,10 +139,12 @@ public class FileReceiver : Gtk.Dialog {
 		});
 	}
 
-	public void set_transfer(string device_name, string device_icon, string path) {
-		device_label.set_markup(GLib.Markup.printf_escaped(_("<b>From</b>: %s"), device_name));
+	public void set_transfer(Bluetooth.Device device, string path) {
+		this.device = device;
+
+		device_label.set_markup(GLib.Markup.printf_escaped(_("<b>From</b>: %s"), device.alias));
 		directory_label.set_markup(Markup.printf_escaped(_("<b>To</b>: %s"), Environment.get_user_special_dir (UserDirectory.DOWNLOAD)));
-		device_image.set_from_gicon(new ThemedIcon(device_icon ?? "bluetooth-active"), Gtk.IconSize.LARGE_TOOLBAR);
+		device_image.set_from_gicon(new ThemedIcon(device.icon ?? "bluetooth-active"), Gtk.IconSize.LARGE_TOOLBAR);
 		start_time = (int) get_real_time();
 
 		try {
@@ -163,22 +166,28 @@ public class FileReceiver : Gtk.Dialog {
 			case "error":
 				notification.set_icon(device_image.gicon);
 				notification.set_title(_("File transfer failed"));
-				notification.set_body(Markup.printf_escaped(_("%s <b>File</b>: %s not received"), device_label.get_label(), transfer.name));
+				notification.set_body(_("File '%s' not received from %s").printf(transfer.name, device.alias));
 				((Gtk.Window) get_toplevel()).application.send_notification("org.buddiesofbudgie.bluetooth", notification);
 				destroy();
 				break;
 			case "queued":
 				break;
 			case "active":
+				// Save the file name here because it won't be available later
+				var name = transfer.filename;
+				if (name != null) {
+					file_name = name;
+				}
+				// Update the transfer progress UI
 				on_transfer_progress(transfer.transferred);
 				break;
 			case "complete":
 				try {
-					move_to_downloads(transfer.filename);
+					move_to_downloads(file_name);
 				} catch (Error e) {
 					notification.set_icon(device_image.gicon);
 					notification.set_title(_("File transfer failed"));
-					notification.set_body(_("File '%s' from %s not received: %s".printf(transfer.name, device_label.get_label(), e.message)));
+					notification.set_body(_("File '%s' from %s not received: %s".printf(transfer.name, device.alias, e.message)));
 					((Gtk.Window) get_toplevel()).application.send_notification("org.buddiesofbudgie.bluetooth", notification);
 					warning("Error saving transferred file: %s", e.message);
 				}
@@ -196,7 +205,7 @@ public class FileReceiver : Gtk.Dialog {
 
 		notification.set_icon(device_image.gicon);
 		notification.set_title(_("File transferred successfully"));
-		notification.set_body(Markup.printf_escaped(_("%s <b>Saved to</b>: %s"), device_label.get_label(), dest.get_path()));
+		notification.set_body(_("Saved file from %s to '%s'").printf(device.alias, dest.get_path()));
 		((Gtk.Window) get_toplevel()).application.send_notification("org.buddiesofbudgie.bluetooth", notification);
 	}
 
--- a/src/dialogs/sendto/Dialog/FileReceiver.vala
+++ b/src/dialogs/sendto/Dialog/FileReceiver.vala
@@ -9,25 +9,12 @@
  * (at your option) any later version.
  */
 
-public class FileReceiver : Gtk.Dialog {
-	public Bluetooth.Device device { get; set; }
+public class FileReceiver : BaseDialog {
 	public string session_path { get; set; }
-
-	public Bluetooth.Obex.Transfer transfer;
-
-	private Gtk.ProgressBar progress_bar;
-	private Gtk.Label device_label;
-	private Gtk.Label directory_label;
-	private Gtk.Label progress_label;
-	private Gtk.Label filename_label;
-	private Gtk.Label rate_label;
-	private Gtk.Image device_image;
 
 	private Notification notification;
 
 	private string file_name = "";
-	private int start_time = 0;
-	private uint64 total_size = 0;
 
 	public FileReceiver(Gtk.Application application) {
 		Object(application: application, resizable: false);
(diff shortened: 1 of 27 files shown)


fossfreedom:
FYI - Debian has now marked the usage of libgnome-bluetooth13 as something to be removed before its next release since upstream has dropped support.

https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1030130 and https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1059486

So great news that we have this PR - nice one @EbonJaeger 

This also dropping support from BCC as well - https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1059485 ...

but with the need to pair devices via BCC that's more problematic - we would have to rewrite this whole section in BCC ... either with the suggested new gnome library ... or a complete rewrite to use bluez?

src/dialogs/sendto/Dialog/FileReceiver.vala
```diff
@@ -0,0 +1,270 @@
...
+		if (seconds < 60 * 60 * 4) {
+			minutes = (seconds - hours * 60 * 60 + 30) / 60;
+			string h = ngettext("%u hour", "%u hours", hours).printf(hours);
+			string m = ngettext("%u minute", "%u minutes", minutes).printf(minutes);
+			///TRANSLATORS: For example "1 hour, 8 minutes".
+			return _("%s, %s").printf(h, m);
+		}
+
+		return ngettext("about %d hour", "about %d hours", hours).printf(hours);
+	}
```
JoshStrobl, line 269: I think it may be worth re-working this function. The initial two conditions are fine, but after that we should probably define minutes as seconds / 60 + a ceil (or a floor, just whether or not we want presentation to be optimistic e.g. report 1 minute when it's technically 1 minute 30s or conservative e.g. report 2 minutes when it might be less, also might play into perceived performance) to drop remainder.

Also curious why we are randomly adding 30s to our seconds.

Once we're at the conditional, we just say if it's minutes < 60, return minutes.

hours + minutes logic seems a bit funky. Am I understanding the conditional after the hours where it's basically if the seconds are less than 4 hours? Any reason for that and not using a modulo to determine total hours then having minutes in the remainder.

src/dialogs/sendto/Dialog/FileSender.vala
JoshStrobl, line 1: Has repeat code to FileReceiver. May be worth some refactoring here.

src/dialogs/sendto/Dialog/ScanDialog.vala
```diff
@@ -0,0 +1,261 @@
...
+			return 1;
+		}
+
+		if (device1.name != null && device2.name == null) {
+			return -1;
+		}
+
+		if (device1.name == null && device2.name != null) {
+			return 1;
+		}
```
JoshStrobl, line 233:
```suggestion
		if (device1.paired && !device2.paired) return -1;

		if (!device1.paired && device2.paired) return 1;

		if (device1.connected && !device2.connected) return -1;

		if (!device1.connected && device2.connected) return 1;

		if (device1.name != null && device2.name == null) return -1;

		if (device1.name == null && device2.name != null) return 1;
```

src/dialogs/sendto/Services/Manager.vala
```diff
@@ -0,0 +1,206 @@
...
+
+			object_manager.object_removed.connect((object) => {
+				object.get_interfaces().foreach((iface) => on_interface_removed(object, iface));
+			});
+		} catch (Error e) {
+			critical("Error getting Bluez object manager: %s", e.message);
+		}
+	}
+
+	// TODO: Do not rely on this when it is possible to do it natively in Vala
```
JoshStrobl, line 57: Let's be real, this could probably be removed when we are almost certainly never going to get around to doing this natively in Vala :D

```suggestion
```

src/dialogs/sendto/Services/Manager.vala
```diff
@@ -0,0 +1,206 @@
...
+	}
+
+	private void register_obex_agentmanager() {
+		try {
+			var connection = GLib.Bus.get_sync(BusType.SESSION);
+			connection.call.begin(
+				"org.bluez.obex",
+				"/org/bluez/obex",
+				"org.bluez.obex.AgentManager1",
+				"RegisterAgent", // TODO: Do we need to worry about unregistering?
```
JoshStrobl, line 84: Does this still need to be done? Is it handled gracefully across panel ~~crashes~~ restarts? That's probably the time I could see it unregistering

src/dialogs/sendto/Services/Manager.vala
```diff
@@ -0,0 +1,206 @@
...
+
+	private void on_interface_removed(DBusObject object, DBusInterface iface) {
+		if (iface is Bluetooth.Device) {
+			device_removed((Bluetooth.Device) iface);
+		} else if (iface is Bluetooth.Adapter) {
+			has_object = !get_adapters().is_empty;
+		}
+	}
+
+	public Gee.LinkedList<Bluetooth.Adapter> get_adapters() requires (object_manager != null) {
```
JoshStrobl, line 118: Bit too late in the Vala land for me to actually make use of this, but TIL about requires.

src/panel/applets/status/BluetoothIndicator.vala
```diff
@@ -12,253 +12,714 @@
...
+	private void update_status() {
+		if (device.connected) {
+			status_label.set_text(_("Connected"));
+			connection_button.show();
+			activatable = false;
+		} else {
+			status_label.set_text(_("Disconnected"));
+			connection_button.hide();
+			activatable = true;
+		}
```
JoshStrobl, line 670: Thoughts?

```suggestion
		activatable = !device.connected;
		status_label.set_text(activatable ? _("Disconnected") : _("Connected"));

		if (activatable) {
			connection_button.hide();
		} else {
			connection_button.show();
		}
```

JoshStrobl (changes requested):
Been using Bluetooth applet and it connects / disconnects just fine with my Fairbuds XL headphones. Battery constantly reports 100% but I'm leaning on that being more an issue with upower than anything else. I'll re-install d-spy and take a look.

Some feedback provided mainly on sendto, some interesting logic for formatting time and duplication that _should_ ideally be avoided.

Design feedback: Maybe could do with some padding around the edges of the popover contents. Night Light and Sound Indicator are a couple examples, but the padding may only really be visible (excluding Night Light header where it's hardcoded into the widget code) when using Materia. Not sure about other themes when they are used in conjunction with the internal theme.

Overall, fantastic work. I can see a lot of care and attention went into this.

JoshStrobl:
Rebased it, ignore noise on the billion commits after my suggestions :D

EbonJaeger:
I was wondering why some of the review items were showing "outdated" xD

src/dialogs/sendto/Services/Manager.vala
EbonJaeger: Yeah, it's pretty nice. If only we'd known about it years ago! xD

src/dialogs/sendto/Dialog/FileSender.vala
EbonJaeger: I just don't know where else the function would live. Do you have any suggestions?

JoshStrobl: Could just be a shared utility class.

JoshStrobl: Or have a class that then builds on Gtk.Dialog to hold the functions and identical props, and extend from there :shrug:

EbonJaeger: > Or have a class that then builds on Gtk.Dialog to hold the functions and identical props, and extend from there 

I ended up doing that. Seems to be the best choice.

EbonJaeger:
The latest commit addresses the duplicated code and estimated time calculation. If you could test receiving a file via Bluetooth, that would be great. I'm not getting the notification to reject or accept the transfer, and I have no idea if it's the code, my phone, or the deity of Bluetooth deciding to mess with me.

JoshStrobl:
@EbonJaeger I am able to send files from my desktop to the test phone just fine but sending from my test phone to the desktop, getting the notification to accept and accepting it, does not work. Just to be sure, I tested it with a unique file not on my desktop.

Also as a note, we need a newline between the From: FILEPATH and "Sent to: DEVICE". Otherwise it all melds into one line.

Will continue testing but hopefully @fossfreedom can do some further testing on his end too since he uses sendto functionality (or at the very least I know he certainly does with more frequency than me -- which is never :D)

---

Otherwise, LGTM. Send-to dialog looked good, though we may want to consider filtering out devices you really can't send to like headphones and wireless keyboards.

fossfreedom:
some quick thoughts.

1. with bluetooth disabled the popover has both a bluetooth settings button in the top section and a big bluetooth button in the popover body.  The latter is perhaps superfluous
2. similarly with nothing paired the second button is there.  
image

3.  the send-to button is enabled even though my phone is marked as disconnected.  Can this be disabled until a bluetooth device is connected?
image

4. 
image

in BCC you click on a connected device and then you send a file directly.
in the popover you click the send files in the top of the popover that then displays this

image

and then you have a whirring discovery option at the top of the dialog and a separate "send" button on the device to send to.

I'm pretty sure this was discussed in #general ... but too far to scroll back - did we discount having a send button next to the disconnect option to directly send to the chosen device - i.e. one click like BCC rather than two clicks via the popover to send a file ?

5. disconnect buttons look a little vertically stretched

image

Tested with Qogir, Arc and Pocillo

6. in that screenshot the second button (Qogir) has a border around the button.  Is this trying to say something specifically about this disconnect button?  Or should both phones disconnect buttons have borders so that the buttons look like buttons?

Tested with a couple of connected phones sending files back and forth.  works nicely

EbonJaeger:
@fossfreedom 

1. & 2. Good point. Removed the button in the placeholder
3. Good idea. Done. (Superseded by below)
4. Ended up being way easier to do than I thought. Devices that support file sending (desktops, laptops, and smartphones; are there any others in [this list](https://www.ampedrftech.com/datasheets/cod_definition.pdf)?) now have their own individual send button that bypasses the first dialog.
5. Still can't get this to happen :/
6. I've noticed that sometimes, buttons don't quite follow their set relief style. Been this way for a while, now.

fossfreedom:
> 4\. Still can't get this to happen :/

hmm - I'm wonder if the send and disconnect buttons should be added to a buttonbox rather than pack-start within a box?

If I connect to a phone via the popover the disconnect button appears in the popover.  Good.  However the send button doesn't appear.  If I restart the panel then both the send button and disconnect buttons appear in the popover.

EbonJaeger:
@fossfreedom I *think* I might have it this time. Can you try the latest commit?

fossfreedom:
Nice work!  Relief issues have been resolved.

Connected my phone - send files button enabled.

Connected my airpods - connected - no send files button - so thats good news.

fossfreedom:
good to be merged IMHO.

PR #468, opened 2023-10-08
Fix initial tray icon scaling

### LINKED ISSUE #467: [Bug] System Tray icon scaling option doesn't work on initial startup
### Description

When first starting the system, the system tray items all act as if scaling is enabled, even though the option is set to disabled in Budgie Desktop Settings. The setting has to be toggled on and off again for it to not scale the icons.

### Budgie version

budgie-desktop 10.8 (git-b3b67fd96b2455d665aa960c950f24259d6b3547)

### Operating System

Solus 4.4 Harmony

1. Log in to the desktop
2. Notice tray icons are large
3. Open Budgie Desktop Settings
4. Go to system tray settings
5. Notice the option to enable scaling is off

### Actual result

Tray items have their icons scaled.

### Expected result

Tray items should not have their icons scaled.

### Additional information

This is occuring for me on my Windowing/Icon Tasklist branch which was rebased to `main` after the option was added.

serebit:
## Description

Resolves #467.

### Submitter Checklist

### DIFF
--- a/src/panel/applets/tray/TrayApplet.vala
+++ b/src/panel/applets/tray/TrayApplet.vala
@@ -207,7 +207,7 @@ public class TrayApplet : Budgie.Applet {
 
 	public override void panel_size_changed(int panel, int icon, int small_icon) {
 		panel_size = panel;
-		items.get_values().foreach((item)=>{
+		items.get_values().foreach((item) => {
 			item.resize(get_target_panel_size());
 		});
 	}
--- a/src/panel/applets/tray/TrayApplet.vala
+++ b/src/panel/applets/tray/TrayApplet.vala
@@ -62,6 +62,7 @@ public class TrayApplet : Budgie.Applet {
 	private uint dbus_identifier;
 	private SnWatcherInterface? watcher = null;
 	private int panel_size;
+	private const int DEFAULT_PANEL_SIZE = 36;
 
 	public TrayApplet(string uuid) {
 		Object(uuid: uuid);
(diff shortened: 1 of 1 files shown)


src/panel/applets/tray/TrayApplet.vala
```diff
@@ -172,7 +172,8 @@ public class TrayApplet : Budgie.Applet {
 		if (key in items) return;
 
 		try {
-			var new_item = new TrayItem(name, object_path, panel_size);
+			var size = settings.get_boolean("scaling") ? panel_size : 36;
```
JoshStrobl, line 175: Given this is duplicated on line 82 during the construction of the TrayApplet, I think we should have this be refactored to a function, and have the `36` be defined as a public const that we can more easily change in one place.

I think there is likely further room for improvement in this area that is out-of-scope of this PR, but maybe worth keeping in mind for the future. I would probably have the panel default size (not user set size, e.g. default panel size defined is 36 already) be used as the basis for the value instead, however that value isn't exposed and we only are informed of panel size changes, so updating applet or applet info to have the associated panel's UUID and make it easier to query those default values would be nice. We're already 99% of the way there in panel.vala when it comes to our from_uuid constructor calls, we'd just want to pass the panel's uuid and have a function. Or maybe we can do a cast of the applet's parent window to a Panel (as Panel inherits from Toplevel, which itself inherits from Gtk.Window) and have a getter live in the panel :shrug:

JoshStrobl (changes requested):

JoshStrobl (approved):
LGTM :+1:

src/panel/applets/tray/TrayApplet.vala
```diff
@@ -199,11 +207,9 @@ public class TrayApplet : Budgie.Applet {
...
 	public override void panel_size_changed(int panel, int icon, int small_icon) {
 		panel_size = panel;
-		if (settings.get_boolean("scaling")) {
-			items.get_values().foreach((item)=>{
-				item.resize(panel);
-			});
-		}
+		items.get_values().foreach((item)=>{
+			item.resize(get_target_panel_size());
+		});
```
EbonJaeger, line 212: Can we add some spaces here to make it less compact?

```suggestion
		items.get_values().foreach((item) => {
			item.resize(get_target_panel_size());
		});
```

EbonJaeger (changes requested):
One style thing, otherwise LGTM!

EbonJaeger (approved):
LGTM, thanks! :)

PR #479, opened 2023-10-15
notifications: Try to guess app id if getting from hints fails

### LINKED ISSUE #477: Notification icons broken [Bug]
### Description

I mentioned this in a comment in [the old issue](https://github.com/BuddiesOfBudgie/budgie-desktop/issues/450), but it didn't seem to get noticed and it wouldn't let me re-open the issue, so I'm posting a new one for visibility.  After updating to 10.8.1, notification icons in raven's notification widget are more broken than before.  More icons do not work properly now and give the default GTK missing icon image.   None of the broken icons that I'm encountering now are custom or referencing something I've modified, so I'm fairly certain it's not a naming issue this time - even nemo's notification about ejecting a USB drive has a broken image.  And many of the icons that are broken are not referencing a file on disk anyway, as they are only data streams sent through dbus. (streamlink-twitch-gui notifying you of a channel coming online is an example.)  The icons used by the actual notification that pops up are fine, this is only in the Raven notification widget.

If it matters, I use the Papirus-Light icon theme, but changing themes doesn't seem to change anything with this bug.

[screenshot](https://i.postimg.cc/HnWv5JSs/notificationbug.png)

### Budgie version

10.8.1

### Operating System

Solus 4.4

Trigger a notification by ejecting a USB drive in nemo or doing various things in lots of apps, then check Raven.

### Actual result

[screenshot](https://i.postimg.cc/HnWv5JSs/notificationbug.png)

### Expected result

correct icon being used

### Additional information

_No response_

EbonJaeger:
## Description

The notification spec states that the `desktop-entry` hint should be the prefix of the Desktop Application Info file. However, because of the push to have fully-qualified app ID's, some applications now have a mismatch between the ID and the desktop file, meaning the hint cannot be used for that application.

In cases where the hint fails, try to guess the ID using the application's name given to us from DBus.

It seems to fix https://github.com/BuddiesOfBudgie/budgie-desktop/issues/477.

### Original description:

The spec states that the image data from a notification should be used before the application icon or any other icon. This change should reflect that.

It seems to fix #477. Nemo and Twitch Streamlink notifications now display properly in Raven on my system.

### Submitter Checklist

### DIFF
--- a/src/lib/notification.vala
+++ b/src/lib/notification.vala
@@ -267,9 +267,13 @@
 
 		private Gtk.Image? get_appinfo_image(Gtk.IconSize size, string? fallback) {
 			if (app_info == null) {
-				var fallback_image = new Gtk.Image.from_icon_name(fallback, size);
-				var invalid_image = (fallback_image == null) || (fallback_image.icon_name == null) || (fallback_image.icon_name == "image-missing") || (fallback_image.icon_name == "");
-				return invalid_image ? null : fallback_image;
+				var theme = Gtk.IconTheme.get_default();
+
+				if (!theme.has_icon(fallback)) {
+					return null;
+				}
+
+				return new Gtk.Image.from_icon_name(fallback, size);
 			}
 
 			var app_icon_name = app_info.get_string("Icon"); // Use the Icon from the respective DesktopAppInfo or fallback to generic applications-internet
--- a/src/lib/notification.vala
+++ b/src/lib/notification.vala
@@ -115,15 +115,27 @@
 				category = variant.get_string();
 			}
 
-			// Set the application ID and app info
+			// Try to set the application ID and app info
 			if ((variant = hints.lookup("desktop-entry")) != null && variant.is_of_type(VariantType.STRING)) {
 				app_id = variant.get_string();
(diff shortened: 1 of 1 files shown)


JoshStrobl:
This change will result in Discord iconography being that of the avatar of the person for every single message, regardless of it being from the individual. That's why I have it set to prioritize the app info image.

JoshStrobl:
See https://github.com/BuddiesOfBudgie/budgie-desktop/pull/452/commits/6c0608b0649cbb7b2faa296c427038800568eb3d

The specification is for the Notification itself, it shouldn't dictate how it should be handled in Raven. I would be curious what sort of AppInfo we are getting from those applications sending the notifications.

EbonJaeger:
Ah, I know why. We stop looking for image data as soon as one is found.

EbonJaeger:
Gonna give this more thought...

JoshStrobl (approved):
Things no bork so LGTM :+1:

EbonJaeger:
Found two more things bork: Solus Update Service and budgie-daemon when we get the notification that there are notifications.

JoshStrobl (approved):

PR #484, opened 2023-10-21
Fix heap buffer overflow in UsageMonitor

serebit:
## Description

The `%ms` format specifier, when passed to POSIX scanf, allocates the necessary space for the value that it reads into a char**.

### Submitter Checklist


EbonJaeger:
How portable is this? It looks like it's an extension done by GNU C, and I don't see any documentation, at least for OpenBSD that supports `%m`.

serebit:
Hm... It's not portable. I'll need to figure out how to preallocate the required space, then.

JoshStrobl:
@serebit Should this be closed?

serebit:
Yeah, I did plan to finish it out eventually but it's not important enough to keep around. If I end up going through Budgie with the sanitizers again I'll open new issues.

PR #485, opened 2023-10-22
Port TabSwitcher, Show Desktop & Workspace applets over to libxfce4windowing

JoshStrobl:
This PR ports the following features over to libxfce4windowing from using libwnck directly:

- Show Desktop applet
- Workspaces Applet
- TabSwitcher

### TabSwitcher

To facilitate tab switcher port and cleanup, considerable amounts of the code was gutted or rewritten, for example wm no longer does any sort of cur_tabs tracking. That is left up to TabSwitcher.

Instead of rebuilding tabswitcher contents every single time we invoke it, we allow it to handle its own state and handle changes based on libxfce4windowing events. Windows are added / removed automatically, workspace handling is done to support show-all-windows function and workspace swapping, and we leverage sort with our recency list and filter to know when to hide items.

### Show Desktop applet

This code implements feature parity with Show Desktop without separately recording any window states. Instead, we just increment over all our stacked windows on the screen and set the minimization state for each window based off the widget toggled state, which gets updated on the screen window_opened and window state changed events.

### Workspaces Applet

Nothing really special to note here about the port. I will however highlight a known issue with Workspace Applet port, which is that windows may not report correct icon. While this is corrected in TabSwitcher with use of AppSystem, I'm going to refrain from duplicating this code into the Workspace applet as I have to then address some issues with AppSystem instantiation triggering segfaults that may otherwise not exist once we centralize it with @EbonJaeger's Windowing namespace / class in his IconTasklist branch, and don't think it's worth adding in more code that'll need to be cleaned up for what is effectively an issue end users won't experience in final release all consumers of similar logic will have been cleaned up by then.

---

## Testing Notes

Be sure to reboot to ensure you are using correct budgie-wm, budgie-daemon, budgie-panel.


src/daemon/tabswitcher.vala
```diff
@@ -83,63 +119,66 @@ namespace Budgie {
...
+		* Construct a new TabSwitcherWindow
 		*/
-		private void on_hide() {
-			var selection = window_box.get_selected_children();
-			Gtk.FlowBoxChild? current = null;
-			if (selection != null && !selection.is_empty()) {
-				current = selection.nth_data(0) as Gtk.FlowBoxChild;
-			}
+		public TabSwitcherWindow() {
+			Object(type: Gtk.WindowType.POPUP, type_hint: Gdk.WindowTypeHint.NOTIFICATION);
```
EbonJaeger, line 146:
```suggestion
		public TabSwitcherWindow() {
			Object(type: Gtk.WindowType.POPUP, type_hint: Gdk.WindowTypeHint.NOTIFICATION);
		}
		
		construct {
```

src/daemon/tabswitcher.vala
```diff
@@ -149,35 +188,120 @@ namespace Budgie {
...
+				if (workspace != null) workspace.activate();
+			} catch (GLib.Error e) {
+				stderr.printf("Failed to activate workspace. Window switching will almost certainly fail: %s\n", e.message);
+			}
+
+			try {
+				tab.window.activate(get_time());
+			} catch (GLib.Error e) {
+				stderr.printf("Failed to activate window: %s\n", e.message);
+			}
```
EbonJaeger, line 292: Activating the window should switch to the correct workspace automatically if it's on a different one. At least, it does with the tasklists and xfce4windowing. I don't recall doing anything special for that.

src/daemon/tabswitcher.vala
```diff
@@ -149,35 +188,120 @@ namespace Budgie {
...
+				libxfce4windowing.Workspace? workspace = tab.window.get_workspace();
+				if (workspace != null) workspace.activate();
+			} catch (GLib.Error e) {
+				stderr.printf("Failed to activate workspace. Window switching will almost certainly fail: %s\n", e.message);
+			}
+
+			try {
+				tab.window.activate(get_time());
+			} catch (GLib.Error e) {
+				stderr.printf("Failed to activate window: %s\n", e.message);
```
EbonJaeger, line 291: Can we not use the GLib logging functions here? I know places in daemon can't for some reason, just making sure this is one of them.

src/daemon/tabswitcher.vala
```diff
@@ -191,43 +315,80 @@ namespace Budgie {
...
+			int64 id_pos = (widget != null) ? get_position_in_recency(widget.id).clamp(0, len - 1) : 0;
+
+			for (var i = 0; i < len; i++) {
+				uint64 new_id_pos = 0;
+
+				if (backwards) {
+					new_id_pos = (id_pos == 0 ? len : id_pos) - (i+1);
+				} else {
+					new_id_pos = id_pos == len - 1 ? 0 : id_pos + (i+1);
+				}
```
EbonJaeger, line 372: Could some comments be added explaining what this is? Maybe I'm just not awake yet, but I have no idea what's happening here.

src/daemon/tabswitcher.vala
```diff
@@ -31,49 +37,79 @@ namespace Budgie {
...
-
-		/**
-		* Last "touched" by the user
-		*/
-		public uint32 usertime;
+		public unowned libxfce4windowing.Window? window = null;
 
-		public unowned Wnck.Window? wnck_window = null;
+		public signal void closed(TabSwitcherWidget widget);
+		public signal void window_activated(libxfce4windowing.Window windiw);
```
EbonJaeger, line 57:
```suggestion
		public signal void window_activated(libxfce4windowing.Window window);
```

EbonJaeger (changes requested):
Looks pretty good for a first look over. One thing I notice is that the tab switcher children don't directly subclass `FlowBoxChild`. I wonder if it would be a good idea to do that; it would avoid having to `get_parent()` and `get_child()` whenever you want to do something to them.

src/daemon/tabswitcher.vala
JoshStrobl: Yep, learned that it automagically does that when working on the Workspace applet. Pretty cool stuff! Can get some code cleaned up here.

src/daemon/tabswitcher.vala
JoshStrobl: Changing it here is probably fine but will check. Sadly it's very easy for me to trigger activation failure (looking at you Visual Studio Code).

src/daemon/tabswitcher.vala
JoshStrobl: Should be sorted as of cdb3a5e

src/daemon/tabswitcher.vala
JoshStrobl: Should be sorted as of cdb3a5e

src/daemon/tabswitcher.vala
JoshStrobl: Should be sorted as of cdb3a5e

JoshStrobl:
@EbonJaeger I addressed your feedback on the idea of sub-classing GtkFlowBoxChild in b08f5fe :+1: Ready for another round of feedback! :grin:

JoshStrobl:
**Known issue with Workspace Applet port:** Windows may not report correct icon. While this is corrected in TabSwitcher with use of AppSysstem, I'm going to refrain from duplicating this code into the Workspace applet as I have to then address some issues with AppSystem instantiation triggering segfaults that may otherwise not exist once we centralize it with @EbonJaeger's Windowing namespace / class in his IconTasklist branch, and don't think it's worth adding in more code that'll need to be cleaned up for what is effectively an issue end users won't experience in final release all consumers of similar logic will have been cleaned up by then.

Going to add this to OP as well.

EbonJaeger (changes requested):
Awesome stuff! Even though it is a bit of a pain having multiple ports in one PR, it would also be a pain to have them separate :stuck_out_tongue:

EbonJaeger (approved):
LGTM, thanks! :)

PR #513, opened 2024-01-28
sendto: Use stateless XDG autostart directory if stateless is set

EbonJaeger:
## Description

Install the sendto autostart file to the correct stateless directory if statelessness has been enabled.

### Submitter Checklist


JoshStrobl (approved):
LGTM, thanks!

Issue #517, opened 2024-01-29
[Bug]  Tab Switcher doesn't work when "Show all windows" is off


Solarunit:
### Description

Tab Switcher doesn't work correctly when "Show all windows in Tab Switcher"  toggle set to **off**
You can't switch windows per workspace independently like it was in 10.8.

### Budgie version

10.9

### Operating System

Arch

### Steps to reproduce the issue

1. set "Show all windows in Tab Switcher" off
2. open some windows on different workspaces, and see how Tab Switcher doesn't work correctly when you trigger it per workspace

### Actual result

it doesn't work

### Expected result

like in 10.8

JoshStrobl:
I'm able to reproduce this in specific scenarios, including re-filtering not occurring on workspace change. Working on a fix.

Issue #515, opened 2024-01-29
[Bug] Alt tab not working properly


durken1:
### Description

When alt tabbing between applications the correct window is displayed but it seems the current active window is reverted back to the last used.

### Budgie version

10.9

### Operating System

EndeavourOS (Linux 6.7.2-arch1-1)

### Steps to reproduce the issue

1. Open a browser
2. Open a terminal emulator 
3. Alt tab back to the browser

### Actual result

The browser shows up, but the terminal is still active. Keyboard shortcuts for the browser does not work and typing results in the text showing up in the terminal.

### Expected result

The browser showing up as well as being the active window, with keyboard shortcuts affecting the browser.

JoshStrobl:
Validated. Working on fix.

PR #518, opened 2024-01-30
fix: multiple tabswitcher bugs (workspace-based rendering, focus, oversizing)

JoshStrobl:
1. Fixed workspace-based rendering by adding missing invalidate filter during construction.
2. Fix focus issue caused by our get_time providing an incorrect event timing. Resolved by using Gdk.X11.get_server_time, need to rip that out when we switch to Wayland.
3. Fix TabSwitcher sizing not updating

Fixes:

- #515
- #517

### DIFF
--- a/src/daemon/tabswitcher.vala
+++ b/src/daemon/tabswitcher.vala
@@ -256,7 +256,7 @@ namespace Budgie {
 			return pos1 < pos2 ? -1 : 1;
 		}
 
-		private uint get_preferred_max_children() {
+		private uint get_visible_children() {
 			uint visible_children = 0;
 			foreach (var child in window_box.get_children()) {
 				var tab = child as TabSwitcherWidget;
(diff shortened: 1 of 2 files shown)


src/daemon/tabswitcher.vala
```diff
@@ -255,6 +256,16 @@ namespace Budgie {
 			return pos1 < pos2 ? -1 : 1;
 		}
 
+		private uint get_preferred_max_children() {
```
EbonJaeger, line 259: The name of this function is misleading. Maybe `get_visible_children` instead, or something?

EbonJaeger (changes requested):
After compiling this, I can no longer get any weirdness to happen with the tab switcher :D

src/daemon/tabswitcher.vala
JoshStrobl: True

JoshStrobl: Resolved with eed33e85a9dd428c8c9a2c6e43ceafc1777abfae

EbonJaeger (approved):
This looks good, and appears to be working well for me. I'm not seeing any more weird behavior with the tab switcher.

JoshStrobl:
@serebit I believe this is something you wanted to test too?

serebit:
Yes, I've installed it and will be testing it later today.

serebit:
Did not investigate workspace applet changes, but both tab switching issues have been fixed to my satisfaction. Approved.

serebit (approved):

PR #525, opened 2024-02-03
Fix workspace applet window icon click not performing workspace switch

JoshStrobl:
Removed unnecessary (and borked) check. Should have been `!==` but not really needed anyways. It won't do anything if the window is already active on the current workspace, so the entire check is extraneous.


EbonJaeger (approved):
LGTM, thanks! :)

PR #540, opened 2024-02-22
Work around even more SNI noncompliance

serebit:
## Description

Fixes Cinny and Tauon tray icons when run outside Flatpak. This cannot fix apps within Flatpak, as their icon paths are only valid within the sandbox.

### Submitter Checklist


EbonJaeger (approved):
LGTM! Tested with Tauon and the indicator shows the app's icon now.

PR #542, opened 2024-02-23
Use Enum DOWN instead of LEFT to scroll in a left-wards direction

### LINKED ISSUE #541: [Bug] Workspace Switcher applet doesn't correctly react to mouse wheel.
### Description

When I put my mouse cursor over the Workspace Switcher applet and turn the mouse wheel down, the workspace switches to the one on the right. OTOH, when I then turn the mouse wheel up (away from me), then nothing happens.

### Budgie version

Budgie 10.9.1

### Operating System

Solus 4.5

1. Change to the workspace on the right.
2. Put the mouse cursor over the Workspace Switcher applet.
3. Turn the mouse wheel up (away from you).

### Actual result

Nothing happens.

### Expected result

The workspace should switch to the one on the left.

### Additional information

_No response_

fossfreedom:
…oses #541)

## Description
Use Enum DOWN instead of LEFT to scroll in a left-wards direction

Closes #541

This is definitely odd.  Looking at libxfce4windowing it does appear that their enum for motion direction is mapped correctly to wnck's values and our vapi is mapped to libxfce4windowing correctly.

So why using DOWN rather than LEFT works feels just wrong ... but works.  Just feels like there is a bug somewhere but not sure where.

### Submitter Checklist

### DIFF
--- a/src/panel/applets/workspaces/WorkspacesApplet.vala
+++ b/src/panel/applets/workspaces/WorkspacesApplet.vala
@@ -207,7 +207,7 @@ namespace Workspaces {
 
 				unowned libxfce4windowing.Workspace current = workspace_group.get_active_workspace();
 				unowned libxfce4windowing.Workspace? next = current.get_neighbor(
-					(down) ? libxfce4windowing.Direction.RIGHT : libxfce4windowing.Direction.LEFT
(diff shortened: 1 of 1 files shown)


JoshStrobl (approved):
LGTM, thanks!

PR #555, opened 2024-04-06
Disable C compiler warnings entirely

serebit:
## Description

In Meson 1.4.0, warnings for Vala-transpiled C code were unhidden, which causes the Budgie build to turn into unreadable logspam due to the absolutely abysmal C code generated by Vala. This PR restores the prior behavior by disabling all warnings for C code.

Let it be known that I didn't want to have to do this and would have fixed the code myself. *If I could have.*

### Submitter Checklist

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -13,10 +13,9 @@ project(
 )
 
 # Vala generates bad C code and missing these on gcc 14 will cause FTBFS
-am_cflags = [
-    '-Wno-incompatible-pointer-types',
-    '-Wno-implicit-function-declaration',
-]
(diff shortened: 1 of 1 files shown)


JoshStrobl (approved):
Cheers!

JoshStrobl:
@serebit Now it triggers the following:

`../src/daemon/tabswitcher.vala:64:38: error: implicit declaration of function ‘xfw_window_x11_get_xid’; did you mean ‘wnck_window_get_xid’? [-Wimplicit-function-declaration]`

So still need those it seems

Issue #565, opened 2024-04-22
[Bug] ITL: Rewrite causes focus issues on window removal


serebit:
### Description

The newly merged rewrite of the ITL causes a reproducible (at least on my machine) issue with window focus. Applications can get into a state where they cannot be focused via the applet, but can be focused by clicking on their window. The inability to focus is cleared upon clicking on the window directly.

### Budgie version

budgie-desktop 10.9.1 (git-ebec0912c0f6db5f8f3c2f1baa6f6d1f67a1cf20)

### Operating System

Arch Linux

### Steps to reproduce the issue

1. Open multiple applications, including one Firefox window
2. Open an additional Firefox window
3. Select a non-Firefox window, then re-select the newly created Firefox window (this sets the non-Firefox window as the next most recent)
4. Close the newly created Firefox window via its menubar Close button (the non-Firefox window should now be focused)
5. Attempt to focus on the remaining Firefox window **via the applet**

### Actual result

https://github.com/BuddiesOfBudgie/budgie-desktop/assets/12981608/0a3d76f6-fe1e-4259-b461-7eb6eda8ea18

### Expected result

On Firefox window removal, Firefox should be focusable via the applet.

PR #566, opened 2024-04-27
windowing: Set a new last active window when the last active window is closed

EbonJaeger:
## Description

Sets the previous active window in a group to another window when a window is closed. This prevents button focus and activation issues when a window in a group is closed.

Fixes https://github.com/BuddiesOfBudgie/budgie-desktop/issues/565

### Submitter Checklist


serebit (approved):
Fixes the issue I was seeing. LGTM!

Issue #424, opened 2023-08-14
[Request] an option for budgie-run-dialog to stay when changing focus


voidpointer0x00:
### Budgie version

budgie-desktop 10.7.2

### Use-cases

I use it to start a set of applications at a time and quite often when I launch an application and start typing the next command/name, one of previous apps appear and the `budgie-run-dialog` is now closed.

So basically **the use case** is sequential launch of an application chain.

### Proposal

 I believe the most convenient way to implement this would be a cli option along the lines of `--keep-focus` or some other. This way it would be possible for the ones interested to reassign the existent `Alf+F2` shortcut to run the same command but with a different option or keep the current behaviour and assign another key combination like `Alf+F3`.

### References

At first I thought https://github.com/BuddiesOfBudgie/budgie-desktop/issues/198 was about the same thing, but it is not :) Although, hopefully, if what I request is not a complicated task, it could be implemented along the side of the referenced issue.

JoshStrobl:
Resolved with #572. Thanks for the contribution @voidpointer0x00, much appreciated!

PR #572, opened 2024-05-10
feat: cli option for focus_quit field #424

voidpointer0x00:
## Description

Added a cli option to disable `focus_quit` that would close `budgie-run-dialog` when loosing focus. Without the argument it's set to `false`, leaving the default behavior unchanged.

```
Application Options:
  -f, --focus-keep     Do not quit when out of focus
```

CLI options parsing was implemented the same way `sendto` `Application.vala` does it.

### Submitter Checklist

### DIFF
--- a/src/dialogs/run/RunDialog.vala
+++ b/src/dialogs/run/RunDialog.vala
@@ -150,6 +150,10 @@ namespace Budgie {
 			);
 		}
 
+		public void set_focus_quit(bool should_quit) {
+			this.focus_quit = should_quit;
+		}
+
 		/**
 		 * Create our launcher buttons from the Budgie AppIndexer.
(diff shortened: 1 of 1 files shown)


JoshStrobl:
Could you explain your usecase as to why you want to keep this open?

voidpointer0x00:
> Could you explain your usecase as to why you want to keep Budgie Run Dialog open?

Yep, it's here https://github.com/BuddiesOfBudgie/budgie-desktop/issues/424

JoshStrobl:
Ah yea sorry, checked this pre-coffee 🤦 

PR looks good to me, thanks!

JoshStrobl (approved):

PR #576, opened 2024-05-18
Fix end session dialog blocking authentication dialog

levkk:
## Description
Fix the end session dialog blocking the authentication dialog when a password is required to shut down the computer.

The fix was to hide the shutdown dialog when the "Shutdown" button is clicked.

### Screenshot
Snapshot_2024-05-18_15-22-47

### Submitter Checklist

### DIFF
--- a/src/daemon/endsession.vala
+++ b/src/daemon/endsession.vala
@@ -66,6 +66,7 @@ namespace Budgie {
 		void restart_clicked() {
 			Closed();
 			ConfirmedReboot();
+			hide();
 		}
 
 		[GtkCallback]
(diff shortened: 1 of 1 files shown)


JoshStrobl:
Thanks for taking the initiative with this PR! While you're at it, could you add `hide()`  calls to the other functions like logout, restart, and move the ` hide()`  call to the end to mirror the cancel function?

levkk:
Logout doesn't require a password prompt as far as I can tell. I added the `hide()` call to the restart function as well. Should be good to go.

JoshStrobl (approved):
Fantastic, thanks!

PR #584, opened 2024-06-03
icons: Use the correct name for the reboot icon

EbonJaeger:
## Description

According to the icon naming spec, the correct name to use for a reboot icon is `system-reboot` (and thus, `system-reboot-symbolic` for symbolic). Before this, we were using `system-restart-symbolic`, and had to provide an icon for this. This change increases compatibility with other icon themes, including Breeze, while retaining our own icon in case a theme does not provide it.

### Submitter Checklist


EbonJaeger:
I would link to the spec, but the Freedesktop Gitlab is currently being upgraded, so it isn't accessible.

JoshStrobl (approved):
LGTM!

PR #587, opened 2024-06-11
icon-tasklist: Fix some pinned launchers not properly updating when opened

EbonJaeger:
## Description

In some cases, opening an application, closing it, and re-opening it again from a pinned launcher would fail to update the launcher button; it would just be stuck doing the waiting animation, despite the application and a window appearing. This is because on X11, you pretty much have to just guess at what the desktop ID is for an application. We do this using the application's name, which may not be in the correct letter casing. This leads to the tasklist searching for, e.g., Nemo.desktop when you really want nemo.desktop.

To get around this, first check using the same ID we did before. If that doesn't work, try again with converting the guessed ID to lowercase.

### Submitter Checklist


JoshStrobl (approved):
I don't really have a way of validating the behavior on X11 anymore so will take your word for it that this works :grin:

PR #588, opened 2024-06-15
feat: implement initial wayland support in panel, panel manager, popovers, and Raven

JoshStrobl:
## Summary

This pull request aims to accomplish the following:

1. Bulk of initial support for Wayland for the Panel, Panel Manager, Popovers, and Raven.
2. Fix annoyances e.g. fullscreen Raven
3. Add workarounds to no-op budgie-wm and IBus management when under Wayland.

**Seriously important:**

Once this is merged, popovers **will** stop working under the X11 environment. Quite likely, more stuff will be broke as well. This **is** intentional, the main branch will be for iterating on Wayland support. If you read this and you aren't a Budgie developer, switch to non-git builds if you're on them. We are not _yet_ providing instructions for running Budgie under Wayland, as it is not ready for broader consumption. You will **not** receive any support running via git until such time, I promise.

## How

The panel, panel manager, and popover management now leverage GtkLayerShell to define anchoring, exclusion zones, and keyboard mode.

Budgie.Popover has been completely dropkicked in favor of Gtk.Popover, as we no longer are limited by the literal window constraints that were imposed in X11 and do not have the seat issues that are present under X11 either.

To workaround issues related to the panel no longer being interactive after moving it via Budgie Desktop Settings, we now have a masked panel recreation (in that we will now hide the panel, update the existing one's placement to ensure struts are updated for everything else, ensure position is saved, nuke and create a new one and show that one).

Some cleanup has happened across relevant parts of the codebase, however PanelManager and Panel itself will need some down the line (probably) even just for the sake of getting rid of gdk warnings related to trying to use APIs that are non-functional on Wayland (mainly related to monitor and positioning).

Other hacks:

- Raven animations because it looked bad. Still weird expansion on show but more usable.
- budgie-wm just doesn't do anything on Wayland, we exit it with status 0. Otherwise :fire:.

## Known Issues

Oh boy, where do I get started?

1. While dock mode works, autohide does not. This should be expected because we have no idea where windows are. This is something the compositor will need to do!
2. Transparency mode untested due to missing protocols in kwin used by libxfce4windowing, same deal as ITL.
3. Keyboard shortcuts do not work. No surprise, no working wm to trigger them.
4. Under kwin, ITL does not update. Expected because lack of protocol support.
5. There is no longer a GtkBox under Popover, therefore no `.container`. We may want to add that stuff back or alternative reach out to any remaining Gtk theme developers to fix theirs.


Conan-Kudo (approved):

Conan-Kudo:
> Under kwin, ITL does not update. Expected because lack of protocol support.

What protocol is missing for the icon task list to work?

JoshStrobl:
> > Under kwin, ITL does not update. Expected because lack of protocol support.
> 
> What protocol is missing for the icon task list to work?

libxfce4windowing requires [wlr foreign toplevel management protocol](https://wayland.app/protocols/wlr-foreign-toplevel-management-unstable-v1) and [ext-workspace-v1](https://gitlab.xfce.org/xfce/libxfce4windowing/-/blob/main/protocols/ext-workspace-v1-20230427.xml?ref_type=heads).

EbonJaeger (approved):
LGTM! Much delete, very wow. Seriously, this all looks good, and I've seen it working live on camera, so hey.

serebit:
Will for sure be reviewing this, but we should release 10.10 before merging so users on the release version can get the fixes and improvements that we've put in.

src/panel/applets/status/PowerIndicator.vala
serebit, line 1: I approve of the formatting changes here, but I would rather they be removed given that this file would be unedited otherwise.

src/panel/applets/status/StatusApplet.vala
serebit, line 1: Same with these changes.

src/plugin/panel/meson.build
```diff
@@ -78,11 +79,11 @@ libpanelplugin_gir = gnome.generate_gir(
     ],
     symbol_prefix: 'budgie',
     identifier_prefix: 'Budgie',
+    header: 'plugin.h',
     install: true,
     extra_args: [
-        '--warn-all',
```
serebit, line 83: Why was this removed?

serebit (commented):
Initial pass, mostly formatting.

JoshStrobl:
> Will for sure be reviewing this, but we should release 10.10 before merging so users on the release version can get the fixes and improvements that we've put in.

Budgie 10.10 is the Wayland release. This has been agreed on in the stand-ups. If we need to cherrypick anything to 10.9.x that can be done via its branch.

Taskfile.yml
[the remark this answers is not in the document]
JoshStrobl: Why? It helps to improve DX by reducing repetition of long content and can be improved with a separate script to handle installation and OS-specific configuration. I believe Solus does something similar to simplify their packaging tooling with go-task as well (@EbonJaeger)

src/plugin/panel/meson.build
JoshStrobl: Good question, will have to double check on that one.

Taskfile.yml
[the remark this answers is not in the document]
serebit: I'm not opposed to adding a taskfile like this, but I don't think it belongs in this PR.

serebit:
> Budgie 10.10 is the Wayland release. This has been agreed on in the stand-ups. If we need to cherrypick anything to 10.9.x that can be done via its branch.

Fair enough.

Taskfile.yml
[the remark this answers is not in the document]
JoshStrobl: That's fair. I'll remove it from this PR and propose it separately 👍 Might give me a good push to sort out the script for configuration too since that is too much of a PITA to do with Taskfile alone

src/panel/applets/status/PowerIndicator.vala
JoshStrobl: Sounds good. Damn format on save :D

src/plugin/panel/meson.build
JoshStrobl: Resolved in d9e2bdef7d3d8bd2cda76b281dc290bfcd305adb

src/panel/applets/status/StatusApplet.vala
JoshStrobl: Resolved in d9e2bdef7d3d8bd2cda76b281dc290bfcd305adb

src/panel/applets/status/PowerIndicator.vala
JoshStrobl: Resolved in d9e2bdef7d3d8bd2cda76b281dc290bfcd305adb

Taskfile.yml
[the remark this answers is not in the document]
JoshStrobl: Resolved in d9e2bdef7d3d8bd2cda76b281dc290bfcd305adb

src/theme/meson.build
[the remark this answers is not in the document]
JoshStrobl: So my logic was that while I was ripping out the popover-manager code that also had checks for older GTK versions, and therefore having to update the theming to account for the new popovers, I would make sure there is alignment on the supported GTK version.

But I evidently got lost in the woods and didn't change the other two instances of meson.build files where 3.22.0 is referenced, so this one indeed looks out of place. Will make a follow-up commit that swaps the others other. Thanks for pointing this one out.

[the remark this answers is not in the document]
JoshStrobl: Correction, it appear I sorted out all of these during development, so gonna close this :D

JoshStrobl:
Upcoming rebase will be to align with a9a9c21b4ffcdef2aed3079ea4f7384cf5101af0 bump to 10.9.2

fossfreedom:
I would recommend we should think about bumping the public gir budgie version - we have always had a pretty stable ABI/API for plugins et al - budgie-1.0.  For wayland though many plugins will simply not work because they have X11 dependencies such as  wnck and gdk-x11.  As the upstream we should be upfront and not just release a new version of budgie that we know 'legacy' plugins will break against because they appear to compile against still.

Lets give an example, I note that third-party plugins that set the popover size no longer run due to "set_default_size" missing as a method on popovers due to the popover refactoring in this PR.  For python plugins, the plugin crashes at runtime - its not obviously detectable that the popover methods have changed.

Distros need to understand that any external plugins they manage need to be relooked at at with budgie-10.10 - bumping the generated gir version number from 1.0 to 2.0 will force this evaluation to see if they work in wayland.

JoshStrobl:
@fossfreedom address with budgie-2.0 bump per the standup, alongside fixing transparent popovers (was due to old reset_style call) and generation of the Night Light applet UI (wrong parent class).

fossfreedom:
Confirmed the budgie bump to v2.0 has had the desired effect of preventing v1.0 applets from being installed.

I note the gir for raven widgets is 1.0 - doubtful this will have any adverse impacts. 

I've merged in these latest changes into my branch and all is well.  Lets crack on with this merge - initial wayland support achieved!

Conan-Kudo (approved):
Let's do it!

PR #595, opened 2024-07-09
Let peas chew on v2.0 for python based plugins

fossfreedom:
## Description
For wayland we need to ensure python based plugins are v2.0

### Submitter Checklist

### DIFF
--- a/src/panel/plugin_manager.vala
+++ b/src/panel/plugin_manager.vala
@@ -34,7 +34,7 @@ namespace Budgie {
 				var repo = GI.Repository.get_default();
 				repo.require("Peas", "1.0", 0);
 				repo.require("PeasGtk", "1.0", 0);
-				repo.require("Budgie", "1.0", 0);
+				repo.require("Budgie", "2.0", 0);
 			} catch (Error e) {
 				message("Error loading typelibs: %s", e.message);
 			}


src/plugin/panel/popover.h
```diff
@@ -16,16 +18,15 @@
 
 G_BEGIN_DECLS
 
-G_DECLARE_DERIVABLE_TYPE(BudgiePopover, budgie_popover, BUDGIE, POPOVER, GtkPopover)
+#define T_TYPE_BUDGIE_POPOVER (budgie_popover_get_type ())
```
EbonJaeger, line 21: By modern conventions, the type should be `BUDGIE_TYPE_POPOVER`.

src/plugin/panel/popover.c
```diff
@@ -11,24 +11,42 @@
...
-static void budgie_popover_class_init(BudgiePopoverClass * c) {
-	(void) c;
+static void budgie_popover_dispose (GObject* object) {
+	BudgiePopover *popover = budgie_popover_get_instance_private(T_BUDGIE_POPOVER (object));
+
+	G_OBJECT_CLASS(budgie_popover_parent_class)->dispose(object);
+}
+
+static void budgie_popover_finalize (GObject* object) {
+	BudgiePopover *popover = budgie_popover_get_instance_private(T_BUDGIE_POPOVER (object));
```
EbonJaeger, line 32: Why is this here?

src/plugin/panel/popover.c
```diff
@@ -11,24 +11,42 @@
...
+
+	G_OBJECT_CLASS(budgie_popover_parent_class)->finalize(object);
 }
 
-GtkWidget * budgie_popover_new(GtkWidget * relative_to) {
-    GtkWidget * popover = g_object_new(BUDGIE_TYPE_POPOVER, "relative-to", relative_to);
-    GtkStyleContext * style = gtk_widget_get_style_context(popover);
-    gtk_style_context_add_class(style, "budgie-popover");
+static void budgie_popover_class_init(BudgiePopoverClass* klass) {
+	GtkWidgetClass* widget_class = GTK_WIDGET_CLASS(klass);
```
EbonJaeger, line 38: This doesn't look like it's being used.

EbonJaeger (changes requested):
Some of this code doesn't look like modern GTK code; I'm not sure how long it's been since that tutorial has been updated.

src/plugin/panel/popover.c
```diff
@@ -11,24 +11,42 @@
...
 #include "popover.h"
 
-struct BudgiePopover {
-    GtkWidget * relative_to;
-};
 
-G_DEFINE_TYPE(BudgiePopover, budgie_popover, GTK_TYPE_POPOVER)
+typedef struct {
+	GtkPopover parent_instance;
+} BudgiePopoverPrivate;
```
EbonJaeger, line 17: This also seems wrong. I would look at [this class](https://github.com/BuddiesOfBudgie/budgie-desktop/blob/main/src/panel/applets/trash/trash_button_bar.c) in the trash applet, which is also a derivable type with private data.

EbonJaeger (changes requested):

PR #627, opened 2024-10-15
Adapt to libxfce4windowing v4.19.8

fossfreedom:
## Description
v4.19.8 includes an API break for the activate method which this PR resolves.

https://gitlab.xfce.org/xfce/libxfce4windowing/-/merge_requests/55

If we find passing NULL isnt adequate then we'll need to fiddle with the Seat changes as per the MR above

Separately I'll raise a PR on main for a larger set of changes once I've tested the PR against magpie v1 and labwc

### Submitter Checklist

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -72,7 +72,7 @@ dep_canberra = dependency('libcanberra')
 dep_canberra_gtk3 = dependency('libcanberra-gtk3')
 
 # Needed for window tracking
-dep_xfce4windowing = dependency('libxfce4windowing-0')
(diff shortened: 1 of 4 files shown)


EbonJaeger (approved):
Now that I have access to the latest XFCE stuff, I have been able to test this. Everything seems to work just fine. I've switched between open windows via the tasklist and the Alt+Tab interface without issue.

JoshStrobl (approved):
LGTM, at the very least it compiles against newest x4w on Fedora rawhide build.

PR #636, opened 2024-11-13
Add our wayland compatible screenlocker

fossfreedom:
## Description
This PR is another "low hanging fruit" - this replaces budgie-screensaver / gnome-screensaver with a wayland compatible screenlocker.

It wires up the various screenlock settings in BCC.

At the same time it wires up the power control settings in BCC to suspend on battery / ac if idle

The how;

it uses swayidle as our idle controller
it uses either gtklock or swaylock for the lockscreen
it uses wlopm to control screen blanking
it uses systemctl to suspend
it uses upower GLIB library to determine battery/ac states.

The power dialog lock button previously was linked to the screensaver lock function - this now has been reorientated to a new dbus entry to perform the same function.

Most definitely this really needs a good testing effort since there are lots of combinations that can be achieved.

gtklock is basically vanilla, but we support distros to either style (gtklock.css) and/or add a config (gtklock.ini) in the DATA for CONFIG budgie-desktop folder (/usr/share/budgie-desktop/ or /etc/budgie-desktop/)

I haven't wired up the BCC "dim" control - this should perhaps be part of Josh's v2 daemon since it needs calling via a wayland client to control the screen brightness.

### Submitter Checklist


src/daemon/screenlock.vala
```diff
@@ -0,0 +1,237 @@
...
+
+            // number of seconds on battery that is inactive before action taken; 0 is never
+            int sleep_inactive_battery_timeout = this.power.get_int("sleep-inactive-battery-timeout");
+
+            // BCC sets "nothing" or "suspend" when toggled
+            string sleep_inactive_battery_type = this.power.get_string("sleep-inactive-battery-type");
+
+            if (sleep_inactive_battery_type == "suspend" && sleep_inactive_battery_timeout != 0) {
+                output = "timeout " + sleep_inactive_battery_timeout.to_string();
+                output += " 'if upower -i $(upower -e | grep '/battery') | grep --color=never -E 'state' | grep -e 'discharging'; then systemctl suspend; fi' ";
```
JoshStrobl, line 66: Related to the upower code, I think we should instead be creating a Up.Client and listening in on the device changes (based on the PowerIndicator.vala code at least seems like changed events call device_added), where we get the DeviceState and determine if the device state is DeviceState.DISCHARGING.

At that point, we would call swayidle if that exists on the system, otherwise use the logind interface similar to UserIndicatorWindow.vala to suspend without interactivity.

Thoughts?

src/daemon/screenlock.vala
```diff
@@ -0,0 +1,237 @@
...
+                Process.spawn_sync ("/",
+							spawn_args,
+							spawn_env,
+							SpawnFlags.SEARCH_PATH |
+                            SpawnFlags.STDERR_TO_DEV_NULL |
+                            SpawnFlags.STDOUT_TO_DEV_NULL,
+							null,
+							null,
+							null,
+							null);
```
JoshStrobl, line 138:
```suggestion
for (i = 0; i <= 1; i++) {
                Process.spawn_sync ("/",
							spawn_args,
							spawn_env,
							SpawnFlags.SEARCH_PATH |
                            SpawnFlags.STDERR_TO_DEV_NULL |
                            SpawnFlags.STDOUT_TO_DEV_NULL,
							null,
							null,
							null,
							null);
}
```

src/daemon/screenlock.vala
fossfreedom: fair enough.  Noted though that PowerIndicator is now using various deprecated APIs.

src/daemon/screenlock.vala
fossfreedom: done

src/daemon/meson.build
```diff
@@ -51,6 +52,7 @@ daemon_deps = [
     dep_xfce4windowing,
     dep_gst,
     dep_cairo,
+    dependency('upower-glib', version: '>= 0.99.8'),
```
serebit, line 55: This dep should be declared in the root meson.build and *referenced* in this file.

serebit (changes requested):

src/daemon/meson.build
fossfreedom: true, true.  Same issue as the PowerInhibitor meson.build file.

We should fix the powerinhibitor deprecation API usage issues and meson.build in a separate PR to not complicate and confuse matters here.

fossfreedom:
> My only bit of feedback would be to make sure we add swayidle and wlopm as dependencies in our meson.build. While _technically_ they are runtime dependencies, it is still a good way to make sure the packager has acknowledged that they are required.

as discussed - new build option to check for runtime dependencies

JoshStrobl (approved):
LGTM :+1: Been testing it for a while now and it's been reliable.

src/daemon/screenlock.vala
```diff
@@ -0,0 +1,312 @@
...
+                    output = "-c 000000";
+                }
+                else {
+                    output = "-i " + current_wallpaper;
+                }
+
+                output = "swaylock -Fklf " + output;
+            }
+
+            if (locker == "gtklock") {
```
serebit, line 59: This isn't a necessary change but rather a suggestion - if the first if condition is true, there's no reason to run the second, so this is a pointless comparison. I would change this to an else if.

src/daemon/screenlock.vala
```diff
@@ -0,0 +1,312 @@
...
+
+                File file = File.new_for_path(current_wallpaper);
+                if (current_wallpaper == "" || !file.query_exists()) {
+                    output = "-c 000000";
+                }
+                else {
+                    output = "-i " + current_wallpaper;
+                }
+
+                output = "swaylock -Fklf " + output;
```
serebit, line 56: Better practice would be to use long-form options so it's obvious to the reader of the code what this is supposed to do, even if it makes the string longer. `-Fklf` is incomprehensible without checking `swaylock --help` - `swaylock --show-failed-attempts --show-keyboard-layout --indicator-caps-lock --daemonize` is completely self-explanatory. I would like to see long-form options used here and elsewhere in this function.

src/daemon/screenlock.vala
```diff
@@ -0,0 +1,312 @@
...
+            }
+
+            if (locker == "gtklock") {
+                output = "gtklock -d";
```
serebit, line 60: Same here (and below with the usages of `-c` and `-s`).

src/daemon/screenlock.vala
```diff
@@ -0,0 +1,312 @@
...
+
+            // number of seconds on battery that is inactive before action taken; 0 is never
+            int sleep_inactive_battery_timeout = this.power.get_int("sleep-inactive-battery-timeout");
+
+            // BCC sets "nothing" or "suspend" when toggled
+            string sleep_inactive_battery_type = this.power.get_string("sleep-inactive-battery-type");
+
+            if (sleep_inactive_battery_type == "suspend" && sleep_inactive_battery_timeout != 0) {
+                output = "timeout " + sleep_inactive_battery_timeout.to_string();
+                output += " 'if dbus-send --print-reply=literal --dest=org.buddiesofbudgie.BudgieScreenlock /org/buddiesofbudgie/Screenlock org.buddiesofbudgie.BudgieScreenlock.OnBattery | grep \"boolean true\" > /dev/null; then systemctl suspend; fi' ";
```
serebit, line 116: Could we use `loginctl suspend` instead? That would allow compatibility with users of elogind on non-systemd systems.

src/daemon/screenlock.vala
```diff
@@ -0,0 +1,312 @@
...
+            }
+
+            if (sleep_inactive_ac_type == "suspend" && sleep_inactive_ac_timeout !=0) {
+                output += " timeout " + sleep_inactive_ac_timeout.to_string();
+                output += " 'if dbus-send --print-reply=literal --dest=org.buddiesofbudgie.BudgieScreenlock /org/buddiesofbudgie/Screenlock org.buddiesofbudgie.BudgieScreenlock.OnBattery | grep \"boolean false\" > /dev/null; then systemctl suspend; fi' ";
```
serebit, line 121: Same here wrt `loginctl`.

src/daemon/screenlock.vala
```diff
@@ -0,0 +1,312 @@
...
+        public bool on_battery() throws GLib.Error {
+            return battery_mode;
+        }
+
+        private void client_daemon(GLib.Object obj, GLib.ParamSpec? sp) {
+            battery_mode = client.get_on_battery();
+        }
+
+        private Screenlock() {
+            string check_apps[] = {"swayidle", "killall", "wlopm", "systemctl", "dbus-send"};
```
serebit, line 236: As above, if we use loginctl we won't be hard-depending on `systemctl` existing, just logind.

serebit (changes requested):
Got some suggestions.

src/daemon/screenlock.vala
fossfreedom: `loginctl suspend` returns unknown command verb 'suspend'.  The man page isnt really revealing here.  Any ideas what the suspend equivalent is?

serebit: Ah, `loginctl suspend` is elogind-only, but there is a dbus method that we can call instead like we do in the power dialogs. See here: https://github.com/BuddiesOfBudgie/budgie-desktop/blob/35f338af03a269652ffb9e0fa2e5203f159ee7ba/src/dialogs/power/dbus.vala#L32

serebit (approved):
LGTM.

PR #638, opened 2024-11-17
Session management

fossfreedom:
## Description
This PR rejigs session management in budgie;

1. the xsession desktop file is now installed in wayland-session; this is the default launching magpie-wm v1
2. Also added in a session for labwc - note this expects labwc as a runtime and we should assume distros are using labwc v0.8.1 or later using wlroots 0.18.x
3. The previous budgie-wm shim has been moved to our power exe.  This allows the popup confirmation to be displayed.  Note this confirmation dialog needs budgie-session
4. We no longer builds budgie-wm - the gnome-session/sessions file no longer autostarts budgie-wm

~Note - this all requires the screenlock branch since the previous budgie-screensaver/gnome-screensaver stuff broke session management.~

### Submitter Checklist


fossfreedom:
> If we are no longer building `wm`, any reason for us to not just nuke it?

There are still bits of capability we have to mull-over - such as OSD support, Alt+Tab comes to mind.  Its easier to have these still in the code base to refer to, grep through etc.  Can nuke it a little later as part of a general cleanup including tidying up manuals and other stuff.

fossfreedom:
Ah... the wm schema keys also need special treatment. Will tidy this

JoshStrobl (approved):
LGTM but all the screenlock bits need to be decoupled from this PR as it is part of another.

fossfreedom:
Rebased on main with a revert of the screenlock commit.

Everything seems to work with the obvious exception of the lock option on the powerdialog - that is the separate PR.

PR #646, opened 2024-12-27
Update it.po

mki77:
## Description
Added and corrected Italian translations.

### Submitter Checklist

### DIFF
--- a/po/it.po
+++ b/po/it.po
@@ -27,23 +27,22 @@ msgstr ""
 #. mode
 #: src/daemon/notifications/dbus.vala:554
 msgid "Unread Notifications"
-msgstr ""
+msgstr "Notifiche non lette"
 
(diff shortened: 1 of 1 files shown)


fossfreedom:
Hi - all translations are managed via transifex.  Please do join and complete the translations.  From time-to-time we pull in the latest translations - there is no need to submit via a PR.

Many thanks

https://explore.transifex.com/buddiesofbudgie/budgie-10/

fossfreedom:
Given that the latest translations are in transifex - they may have already changed compared to what is already in github.

Thus we cannot merge the PR here because it may be behind what is already in transifex.

JoshStrobl:
You'd have to be more specific on what specific dropdowns you are referring to.

JoshStrobl:
No worries, thank you for raising it and I appreciate all the work you have done on the translations :heart:

PR #647, opened 2025-01-01
Change our Desktop Current Session ID to Budgie from Budgie:GNOME

fossfreedom:
## Description
This changes our XDG_CURRENT_DESKTOP identifier from Budgie:GNOME to Budgie.

i.e. commit https://github.com/BuddiesOfBudgie/budgie-desktop/pull/647/commits/669cc032547d8b835bb18a55f745f461aae8695e

This symbolically further unties Budgie from its GNOME based roots.

We have the opportunity here to strip some of the settings-daemon plugins that is no longer valid with our initial wayland journey.  MediaKeys & Keyboard has specifically been dropped - side effect is that these gsd_ based executables no longer spamming journald because they were expecting a mutter backend.

Other side effects will be any autostart previously started with "GNOME" no longer autostarts.

Marking this as draft.  This PR will need to be rebased once the session PR is merged.

### Submitter Checklist


JoshStrobl (approved):
LGTM :+1:

PR #651, opened 2025-01-05
Screen Dimming

fossfreedom:
## Description
This PR adds power-dimming support if configured in BCC Power 

i.e. if idle-dim is configured then swayidle  will dim after 30 seconds and resume with the original set brightness.
It uses gsd-power to configure brightness settings.

Marking as draft until the screenlock PR is  dealt with and then this will need rebasing.

### Submitter Checklist


src/daemon/screenlock.vala
```diff
@@ -0,0 +1,375 @@
...
+
+                /**
+                * Try in order, and load the first one that exists:
+                * - /etc/budgie-desktop/[gtklock.ini | gtklock.css]
+                * - /usr/share/budgie-desktop/[gtklock.ini | gtklock.css]
+                */
+                string[] system_configs = {
+                    @"file://$(Budgie.CONFDIR)/budgie-desktop/gtklock.ini",
+                    @"file://$(Budgie.DATADIR)/budgie-desktop/gtklock.ini"
+                };
```
EbonJaeger, line 83: Should we care about files in `~/.config`?

src/daemon/screenlock.vala
```diff
@@ -0,0 +1,375 @@
...
+            return output;
+        }
+
+        private string calculate_dim() {
+            isdimmable = this.power.get_boolean("idle-dim");
+
+            if (!isdimmable) return "";
+
+            return "timeout 30 'dbus-send --type=method_call --dest=org.buddiesofbudgie.BudgieScreenlock /org/buddiesofbudgie/Screenlock org.buddiesofbudgie.BudgieScreenlock.Dim' resume 'dbus-send --type=method_call --dest=org.buddiesofbudgie.BudgieScreenlock /org/buddiesofbudgie/Screenlock org.buddiesofbudgie.BudgieScreenlock.Undim' ";
+        }
```
EbonJaeger, line 146: Why do this when we have the methods available to us right in this class?

src/daemon/screenlock.vala
```diff
@@ -0,0 +1,375 @@
...
+				conn.register_object(DBUS_SCREENLOCK_PATH, this);
+			} catch (Error e) {
+				message("Unable to register Screenlock: %s", e.message);
+			}
+		}
+
+        public async void lock() throws GLib.DBusError, GLib.IOError {
+            if (!all_apps) {
+                yield;
+                return;
```
EbonJaeger, line 239: Do we have to yield here?

src/daemon/screenlock.vala
```diff
@@ -0,0 +1,375 @@
...
+				print("Error: %s\n", e.message);
+			}
+
+            yield;
+        }
+
+        public async void dim() throws GLib.DBusError, GLib.IOError {
+            if (powerscreen_proxy == null) {
+                yield;
+                return;
```
EbonJaeger, line 254: Also here?

EbonJaeger (changes requested):
There is an interesting mix of spaces and tabs used for indentation in the `screenlock.vala` file, in some cases, both on the same line.

src/daemon/screenlock.vala
fossfreedom: 'fraid I don't understand the comment.  Here we are adding a timeout/resume to swayidle to perform the dim/undim stuff if its set in BCC to screendim

src/daemon/screenlock.vala
fossfreedom: Fair enough - added in the latest commit

fossfreedom:
> There is an interesting mix of spaces and tabs used for indentation in the `screenlock.vala` file, in some cases, both on the same line.

That was resolved via the original screenlock PR - now merged.

src/daemon/screenlock.vala
```diff
@@ -68,16 +90,19 @@ namespace Budgie {
...
+				string[] lock_configs = {
+					"file://" + Environment.get_user_config_dir() + "/budgie-desktop/gtklock.ini",
 					@"file://$(Budgie.CONFDIR)/budgie-desktop/gtklock.ini",
 					@"file://$(Budgie.DATADIR)/budgie-desktop/gtklock.ini"
 				};
 
-				foreach (string? filepath in system_configs) {
+				foreach (string? filepath in lock_configs) {
 					File file = File.new_for_uri(filepath);
+					warning(filepath);
```
JoshStrobl, line 105: Remove warning. Guessing it was there when you were debugging the addition of the gtklock.css from user config dir.

```suggestion
```

src/daemon/screenlock.vala
EbonJaeger: After going through this with Josh, I now understand that this is passed as an argument to `swayidle`. That bit wasn't clear to me, so I thought this was trying to call our own functions via dbus-send from our own binary.

EbonJaeger (approved):
Do it up.

PR #654, opened 2025-01-16
Add a root menu for the desktop

fossfreedom:
## Description
This PR adds a root menu where no desktop icons is used.

The bridge has been updated to copy the menu from the installed shared folder - or from the distro- equivalent if a distro wants their own custom folder.

Translations for the menu labels are handled as well on the startup of the bridge.

Note - the old menu code will be deleted in a separate PR in-coming via the osdkeys branch

### Submitter Checklist


JoshStrobl (approved):
LGTM :+1:

PR #656, opened 2025-01-18
Rejig notifications to work with layer-shell positioning

fossfreedom:
## Description
This PR changes from absolute positioning of windows (aka X11) to layershell positioning

### Submitter Checklist

### DIFF
--- a/src/bridges/labwc/rc.xml
+++ b/src/bridges/labwc/rc.xml
@@ -300,11 +300,6 @@
 		<windowRule identifier="nemo-desktop" title="nemo-desktop" skipTaskbar="yes" skipWindowSwitcher="yes">
 			<action name="ToggleAlwaysOnBottom" />
 		</windowRule>
-		<windowRule identifier="budgie-daemon" title="BudgieNotification" skipTaskbar="yes" skipWindowSwitcher="yes">
-			<action name="SetDecorations" decorations="None" />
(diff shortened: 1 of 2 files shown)


EbonJaeger:
Could we use anchors to set the corner of the screen selected in Budgie Desktop Settings?

fossfreedom:
@EbonJaeger hopefully the last commit does this.  Needs proper testing rather than just "notify-send" stuff that I have been using.

EbonJaeger:
Using a margin for y-axis positioning is really clever; I like it. In the calculate position function, do we still have switch cases for left and right? If so, it could likely be further simplified, since we only care about the y-axis. Left side or right side doesn't matter anymore.

fossfreedom:
> Using a margin for y-axis positioning is really clever; I like it. In the calculate position function, do we still have switch cases for left and right? If so, it could likely be further simplified, since we only care about the y-axis. Left side or right side doesn't matter anymore.

indeed - latest commit does this.

EbonJaeger:
Since the calculate function only has to return one value now, should we make the return type `int` and remove the `out` parameter variable?

EbonJaeger (approved):
I love seeing that function gutted :smiley: 

It appears to work great for me; I sent some notifications with different position settings, and it just works. No more notifications in the middle of the screen!

JoshStrobl (approved):
LGTM 👍

PR #657, opened 2025-01-20
OSD capability

fossfreedom:
## Description
This PR reworks the previous mutter invoked OSD capability to be wayland friendly.

Now we specifically look for volume & brightness changes are display the OSD.
The volume changes are tracked in the same manner as the volume applet - i.e. GVC.  
For brightness we grab the values straight out of the gsd component.

New here is OSD display when the caps-lock / num-lock keys are displayed.  Very useful for small keyboards when you can easily press the caps-lock key accidentally.  Note the comment in the keymap section of osdkeys where there appears to be a bug under wayland on first initialisation of the keymap - tests on X11 show the this code works just fine.  i.e. if you press caps lock immediately on first login the OSD is not displayed. Quick test under Gtk4 and the equivalent standalone code for keymap initialisation works just fine.  ho hum.

The old menus code has been nuked.

The daemon X11 stuff has also been removed - budgie-daemon is now running wayland not xwayland.

### Submitter Checklist


EbonJaeger:
I wonder if that's what caused a thread to be constantly at 100% the other day... I'd been trying to figure out what caused it, but couldn't get it to happen again.

fossfreedom:
@BuddiesOfBudgie/best-buds ok - this is ready.  The big issue I found is the way we are using libxfce4windowing.

If you either:

1. call libxfce4windowing.Screen.get_default() too quickly and/or
2. keep a reference such as a class variable to libxfce4windowing.Screen.get_default() and subsequently call the method somewhere else

Then its very very likely that budgie-daemon will end up spinning indefinitely at 100%.  Running a gdb -p on budgie-daemon shows that the callout to the underlying wayland-client stuff (libwayland) performs a mutex-lock - two calls cause each-other to block and never completes.

Thus - we should limit our usage of libxfce4windowing calls throughout a process especially if the usage is impacted as described by 1/2 above.

src/daemon/osdkeys.vala
```diff
@@ -0,0 +1,254 @@
...
+			mixer.open();
+
+			// wait a short while to allow the async mixer methods to complete otherwise
+			// we see the volume OSD on startup accidently when keymap changes are invoked
+			Timeout.add(200, () => {
+				initialising = false;
+				map = Gdk.Keymap.get_for_display(Gdk.Display.get_default());
+				map.state_changed.connect(on_keymap_state_changed);
+				return false;
+			});
```
EbonJaeger, line 84: Could we do this in the `on_mixer_state_change` handler once the mixer says it's ready?

src/daemon/osdkeys.vala
```diff
@@ -0,0 +1,254 @@
...
+		  state of the capslock/numlock is not set until another keymap is activated
+		  This can be ctrl/shift as well as the caps/numlock. If the latter is activated
+		  this can mean the OSD for keymap state is not activated on first use; only on
+		  subsequent keypresses.
+		 */
+		private void on_keymap_state_changed() {
+			if (!firstrun) {
+				capslock = map.get_caps_lock_state();
+				numlock = map.get_num_lock_state();
+				firstrun = true;
```
EbonJaeger, line 173: This feels backwards. Doesn't this actually happen on the first run?

EbonJaeger (changes requested):
Overall this looks good. I just have a couple of things.

src/daemon/osdkeys.vala
```diff
@@ -0,0 +1,254 @@
...
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2 of the License, or
+ * (at your option) any later version.
+ */
+
+ namespace Budgie {
+
+	[DBus (name = "org.budgie_desktop.BudgieOSD")]
+	interface ShowOSD : GLib.Object {
+		public abstract async void ShowOSD(HashTable<string,Variant> params) throws DBusError, IOError;
```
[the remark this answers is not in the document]
fossfreedom: happy to

src/daemon/osdkeys.vala
```diff
@@ -0,0 +1,254 @@
...
+		private bool initialising;
+
+		public OSDKeys() {
+			initialising = true;
+			try {
+				osd = Bus.get_proxy_sync (BusType.SESSION, "org.budgie_desktop.BudgieOSD",
+															"/org/budgie_desktop/BudgieOSD");
+
+			} catch (Error e) {
+				stderr.printf ("%s\n", e.message);
```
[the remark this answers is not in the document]
fossfreedom: absolutely

src/daemon/osdkeys.vala
fossfreedom: As per the comment above - 'fraid not.  It appears GTK3 is confused under wayland with states and it needs the user to hit ctrl/shift/caps/numlock for it to update the states correctly.

src/daemon/osdkeys.vala
fossfreedom: 'fraid not.  Anything, whether the panel volume controller or any autostart file that touches the volume stuff can poke at the mixer to show the volume OSD.  

The slight delay here mitigates this effect.

EbonJaeger (approved):
LGTM, thanks!

PR #660, opened 2025-01-22
meson: Allow custom prefixes for third-party packages

bobby285271:
## Description

labwc, magpie, gsd does not always share the same prefix as budgie-desktop and sometimes they are just installed in some unusual paths. For example:

- In NixOS labwc is installed in `/nix/store/xxx-labwc-0.8.2/bin`.
- In Arch Linux gsd-media-keys is installed in `/usr/lib/` ([reference](https://gitlab.archlinux.org/archlinux/packaging/packages/meson/-/blob/1.6.1-1/arch-meson?ref_type=tags#L6)).

Simply don't hardcode these paths.

### Submitter Checklist


fossfreedom (approved):
Really appreciate the effort here.  Makes total sense. Cheers!

PR #661, opened 2025-01-23
Switch session files, leave magpie session in the code for a future release

fossfreedom:
## Description
Switch the sessions to make labwc the default for the 10.10 release.

Leave the magpie session in the code, but commented out.  We can later reinstate this as a default as and when needed.

### Submitter Checklist


JoshStrobl (approved):
LGTM :+1:

PR #662, opened 2025-01-24
icon-tasklist: Remove last references to xid

EbonJaeger:
## Description

Removes the last remaining reference to XID in Icon Tasklist.

### Submitter Checklist

### DIFF
--- a/src/panel/applets/icon-tasklist/widgets/ButtonPopover.vala
+++ b/src/panel/applets/icon-tasklist/widgets/ButtonPopover.vala
@@ -241,7 +241,6 @@ public class ButtonPopover : Gtk.Popover {
 private class WindowControls : Gtk.Box {
 	public libxfce4windowing.Window window { get; construct; }
 
-	private Gtk.CheckButton? keep_on_top_button;
 	private Gtk.Button? maximize_button;
 	private Gtk.Button? minimize_button;
 	private Gtk.Button? return_button;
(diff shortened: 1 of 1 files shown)


src/panel/applets/icon-tasklist/widgets/ButtonPopover.vala
```diff
@@ -257,6 +257,11 @@ private class WindowControls : Gtk.Box {
 			relief = Gtk.ReliefStyle.NONE,
 		};
 
+		// Keep on top is not currently supported on Wayland
+		if (libxfce4windowing.windowing_get() == libxfce4windowing.Windowing.WAYLAND) {
+			keep_on_top_button.sensitive = false;
+		}
```
JoshStrobl, line 263: Given we aren't going to support X11, I think it'd be better just to remove the functionality entirely, as 10.10+ this button would just be permanently disabled.

JoshStrobl (changes requested):

src/panel/applets/icon-tasklist/widgets/ButtonPopover.vala
EbonJaeger: It should be possible to support always-on-top on Wayland, but it has to be implemented by libxfce4windowing. Using labwc, I can set a window to always be on top from the right click menu on a window's titlebar. I'm not sure which protocol is used for this; maybe [`wlr-layer-shell`](https://gitlab.freedesktop.org/wlroots/wlroots/-/blob/master/protocol/wlr-layer-shell-unstable-v1.xml?ref_type=heads)?

JoshStrobl (approved):
LGTM, thanks! 👍

PR #667, opened 2025-01-25
Abomination is no longer used since we started using libxfce4windowing

EbonJaeger:
## Description

We no longer use the old Abomination code since Icon Tasklist has been ported to use `libxfce4windowing`.

### Submitter Checklist


fossfreedom (approved):
LGTM!

PR #668, opened 2025-01-25
Add on-demand to the polkit window to ensure entry field is editable

fossfreedom:
## Description
This ensures the polkit window editable field is focused.

Note - it appears that wlroots based compositors need the user to run the following to allow polkit windows to have the permissions to actually execute what pkexec wants to run for xwayland apps.

    xhost + local:

Without the above in journalctl you see the following:

    Authorization required, but no authorization protocol specified

There maybe some other secret sauce here get the extant polkit code to-do this for the user, but I don't know what this wouldbe.

### Submitter Checklist


EbonJaeger (approved):
LGTM, thanks!

PR #672, opened 2025-02-06
Add nightlight support using gammastep - closes #649

### LINKED ISSUE #649: Implement Wayland support in Night Light applet
Night Light needs to be ported to [wlr-gamma-control-unstable-v1](https://wayland.app/protocols/wlr-gamma-control-unstable-v1) / budgie-daemon v2 / gammastep (as a quick hack). 

The easiest route would be adding gammastep as a runtime dependency and starting / stopping gammastep with a one-off gamma set for our temperature value. Schedule is a bit trickier due to geoclue + MLS (Mozilla Location Services) being non-functional, so we would need to provide a setting in Budgie Control Center for setting your lat / lng instead.

Afterwards, we would port it to using budgie-daemon v2 when it supports its own SetGamma or SetTemperature on an output configuration object. This is expected to be implemented after the configuration batch system. At that point, the Night Light applet would just create new configuration objects to be applied that'd toggle it on / off or set temperature.

fossfreedom:
## Description
Adds nightlight support using gammastep.

budgie-daemon watches for settings-daemon.color changes.  On any change, a gammastep.config file is used to define what values should be used.

This config files exists in /usr/share/budgie-desktop and like the labwc bridge, config files can be overridden by a distro- prefix to the config file name.

By using a config file, all gammastep configuration options can be modified by a user - with the exception of the key-pairs managed directly by the daemon. 

### Submitter Checklist

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -73,6 +73,7 @@ cdata = configuration_data()
 if get_option('with-runtime-dependencies')
     find_program('wlopm', required: true)
     find_program('swayidle', required: true)
+    find_program('gammastep', required: true)
     found_gtklock = find_program('gtklock', required: false)
     if found_gtklock.found() == false
         find_program('swaylock', required: true)
--- a/src/daemon/gammastep.config
+++ b/src/daemon/gammastep.config
@@ -32,7 +32,7 @@ fade=1
 #brightness-night=0.4
 # Set the screen gamma (for all colors, or each color channel
 # individually)
-gamma=0.8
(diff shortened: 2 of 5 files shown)


fossfreedom:
^^^ that build failure appears to be rawhide specific.

EbonJaeger:
This apparently needs a rebase again.

EbonJaeger (approved):
LGTM, let's do it.

PR #674, opened 2025-02-12
fix: workspace applet wayland support

### LINKED ISSUE #470: Implement Wayland support in Workspaces applet
This is a meta-tasklist for implementation of Wayland support in Workspace Applet, such as via use of [libxfce4windowing](https://gitlab.xfce.org/xfce/libxfce4windowing). Strict usage of Wnck is not viable as that is limited to X11. libxfce4windowing is designed to support both X11 and Wayland.

EbonJaeger:
## Description

This PR does a few things for the workspace applet:

- Removes any reference to BudgieWM, since that is no longer used.
- Removes any remaining Wnck/xid code.
- Fixes a crash when a window is opened.
- Makes the add workspace button unlickable if a new workspace cannot be added.

The only other outstanding issue that I know of so far is that hovering the mouse over the add workspace button when it's set to only show on hover causes the revealer to keep opening and closing in rapid succession for as long as the mouse is there.

This should fix #470. If any further refinements are needed, they can be done later in separate PRs.

### Submitter Checklist

### DIFF
--- a/src/panel/applets/workspaces/WindowIcon.vala
+++ b/src/panel/applets/workspaces/WindowIcon.vala
@@ -11,19 +11,30 @@
 
 namespace Workspaces {
 	public const int WORKSPACE_ICON_SIZE = 16;
+	public const string FALLBACK_ICON_NAME = "image-missing";
 
 	public class WindowIcon : Gtk.Button {
-		private libxfce4windowing.Window window;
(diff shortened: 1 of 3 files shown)


fossfreedom (approved):
As you say - key is that it fixes the crash.  Its actually very usable.  Lets get this in and iterate on it with later changes.

Nicely done!

PR #676, opened 2025-02-23
Test to see if the gammastep local folder actually exists before creating it

fossfreedom:
## Description
nightlight copies a default config from a system location to a local folder.

Normally this is ok on a first login.  

However, this local folder may already exist if gammastep is in use from another DE.

So lets test to see if the folder actually exists before trying to create the gammastep folder... thus allowing the budgie config to be successfully copied into the gammastep folder

### Submitter Checklist

### DIFF
--- a/src/daemon/nightlight.vala
+++ b/src/daemon/nightlight.vala
@@ -74,11 +74,16 @@ namespace Budgie {
 
 			if (default_config_path != null && default_config_path != local_config_path) {
 				try {
-					// Copy the default configuration file to the local configuration file location
+					// Create the local folder to hold our configuration file
 					File dir = File.new_for_path(Path.get_dirname(local_config_path));
 					if (!dir.query_exists()) {
 						dir.make_directory_with_parents(null);
 					}
+				} catch (Error e) {
+					warning("Failed to create local folder: %s\n", e.message);
+				}
+				try {
+					// Copy the default configuration file to the local configuration file location
 					File default_config_file = File.new_for_path(default_config_path);
 					File local_config_file = File.new_for_path(local_config_path);
 					default_config_file.copy(local_config_file, FileCopyFlags.NONE, null, null);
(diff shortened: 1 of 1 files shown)


EbonJaeger:
For the sake of having atomic file operations, could we:

1. Wrap the directory create call in a try/catch
2. In the `catch` block, check the error and if:
  i. The error is directory already exists, ignore the error and continue
  ii. The error is any other error, handle it appropriately (like logging a critical and returning)

This way, we never have to worry about the file system changing between us querying it and us creating it.

fossfreedom:
make sense. done.

EbonJaeger (changes requested):
Closer to what I was after. What I mean is, in the new `try` block, don't query if the directory exists at all; just try to create it. Then, in the `check` block, check if the error code (`e.code`) is equal to `FileError.EXISTS`. If it is, print nothing and continue on, otherwise print the error and return.

src/daemon/nightlight.vala
```diff
@@ -74,9 +74,17 @@ namespace Budgie {
...
 			if (default_config_path != null && default_config_path != local_config_path) {
 				try {
-					// Copy the default configuration file to the local configuration file location
+					// Create the local folder to hold our configuration file
 					File dir = File.new_for_path(Path.get_dirname(local_config_path));
 					dir.make_directory_with_parents(null);
+				} catch (IOError.EXISTS exist_error) {
+					/* It's not an error to worry about if the folder exists */
+				} catch (Error e) {
+					warning("Failed to create local folder: %s\n", e.message);
```
EbonJaeger, line 83: Do we need to exit here, if we failed to create the directory and it doesn't exist?

fossfreedom: Well, gammastep will still run, just not use any custom values set. The alternative is not having gammastep running at all... and then a user has to figure out why nothing appears to work when switching on/off.

EbonJaeger: Hmm. So, our choices are either Gammastep runs with potentially unexpected settings, or it doesn't run at all. Neither is very great, at least not without somehow telling the user. Let's think about that later, though.

EbonJaeger (approved):
LGTM, thanks!

PR #680, opened 2025-03-04
Resolve vala 0.56.18 compact class inheritance removal - closes #679

### LINKED ISSUE #679: [Bug] FTBFS with Vala 0.56.18
On budgie-desktop main (commit 7e7ed5d5c6df42751758ecb6082278ca78f1c323), when compiling with Vala 0.56.18, compilation fails due to the following error:

```
../src/lib/animation.vala:31.2-31.37: error: Compact class `Budgie.Animation' cannot inherit from non-compact class `GLib.Object'
   31 | 	public class Animation : GLib.Object {
      | 	^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~  
```

I've confirmed that this error doesn't happen on 0.56.17.

fossfreedom:
## Description
With a shrug for the vala microrelease introducing a breaking change, the PR simply removes the use of [Compact] where inheritance issues breaks compilation

### Submitter Checklist


fossfreedom:
sight ^^^ fedora breakage again - nothing to-do with this PR

EbonJaeger:
The question is if we want to use compact classes or not. At least for `Animation`, a compact class may actually make sense. In which case, could we just remove the `: GLib.Object` part of the definition?

fossfreedom:
Possibly.  Quick test though of removing glib and making the private member to public because that's apparently how stuff is now supposed to work.... raven crashes on opening.

Icon task list I think works but given what I saw for raven, not sure about absolute stability.

serebit (approved):
Compiles on my system. Would like someone to make sure the changes work properly outside of Arch, but LGTM.

fossfreedom:
> Would like someone to make sure the changes work properly outside of Arch

A variation of this patch (for 10.9.2) is in both Debian and Ubuntu 25.04 since they now ship this vala version.

JoshStrobl:
I'll get this backported for Fedora as well since they decided now would be a good time to land it with a release right around the corner.

fossfreedom:
@JoshStrobl in that case a PR on the 10.9.2 branch would be warranted as well?

JoshStrobl:
I can cherrypick it to branch once it's merged, no sense in another PR imo

fossfreedom:
The 10.9.2 has an extra source file that has "compact" attached to it - that source file has been refactored out in main.

JoshStrobl:
Ah yea duplicate animation class in icontasklist. Feel free to just apply that patch directly in the branch, I don't think a review will be necessary.

PR #683, opened 2025-03-14
Add override for overlay-key to prevent crash with mutter-common v48-rc

fossfreedom:
## Description
Use  the pre- overlay-key value to prevent magpie window manager crash when mutter-common v48-rc is installed

### Submitter Checklist


JoshStrobl:
Thanks!

PR #685, opened 2025-03-15
Bump the plugin version since the API/ABI has been updated

fossfreedom:
## Description
This is required to prevent old plugins crashing the panel due to API/ABI changes

### Submitter Checklist


JoshStrobl (approved):
LGTM :rocket:

PR #687, opened 2025-03-19
Removal modal dialog and unredirect support since these are mutter specific

fossfreedom:
## Description
as per the title - BDS - Windows options are not relevant for 10.10

### Submitter Checklist


JoshStrobl (approved):
LGTM!

PR #689, opened 2025-03-22
Cleanup

fossfreedom:
## Description
Fixes the build of the trash applet on chroot based installs

General tidy-ups of meson vapi section - deduplicates specification of packages that have already been declared as a dependency in the same file.

### Submitter Checklist


src/panel/applets/trash/trash_applet.c
```diff
@@ -17,6 +17,7 @@
 #include "trash_notify.h"
 #include "trash_popover.h"
 #include "trash_settings.h"
+#include <budgie-enums.h>
```
EbonJaeger, line 20: This doesn't feel like the correct header. It should be `popover.h`, no?

fossfreedom: Fair enough - compilation works either way.  So pushed that instead.

Still doesnt though fix the underlying issue of src/windowing not being compiled correctly into the icon-tasklist :(

src/panel/applets/trash/trash_applet.c
```diff
@@ -17,6 +17,7 @@
 #include "trash_notify.h"
 #include "trash_popover.h"
 #include "trash_settings.h"
+#include <popover.h>
```
EbonJaeger, line 20: I think we might want this to be:

```c
#include "popover.h
```

I think by using brackets, it will point to the system version of the header, whereas we probably want to use the current version in the source tree. @JoshStrobl am I correct in this assumption?

EbonJaeger (approved):
LGTM, thanks!

PR #690, opened 2025-03-24
use budgie friendly folder for our labwc config file

fossfreedom:
## Description

as per xfce4-session they use a config file specific to xfce4 i.e. ~/.config/xfce4/labwc

Lets be labwc friendly and similarly use a DE specific folder i.e. ~/.config/budgie-desktop/labwc

At the same time fix a crash on first startup where we were trying to translate the system menu.xml file (causing a permissions failure) rather than the freshly copied local menu.xml which can be written to.

### Submitter Checklist


EbonJaeger (approved):
LGTM, thanks!

PR #691, opened 2025-03-29
Use a static shared-library for the windowing capability to ensure chroot based packages have something to link against

fossfreedom:
## Description
Similarly for other code library usage (plugins, themes, panels etc), this defines a shared library for the windowing codebase to allow Debian (and possibly others) something to link against. 

The effect of this change is to have a functional icon-tasklist with new windows appearing as icons.

### Submitter Checklist

### DIFF
--- a/src/windowing/meson.build
+++ b/src/windowing/meson.build
@@ -9,24 +9,35 @@ libwindowing_deps = [
   dep_xfce4windowing,
 ]
 
-libwindowing = static_library(
-  'windowing',
+libwindowing = shared_library(
+  'budgie-windowing',
   libwindowing_sources,
   dependencies: libwindowing_deps,
   vala_args: [
     '--pkg', 'gio-unix-2.0',
     '--pkg', 'gtk+-3.0',
     '--pkg', 'libxfce4windowing-0',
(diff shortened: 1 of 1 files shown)


src/windowing/meson.build
```diff
@@ -9,24 +9,35 @@ libwindowing_deps = [
...
+)
+
+# Allow building against libbudgie-windowing
+pkgconfig.generate(
+    name: 'Budgie Windowing',
+    description: 'Budgie Windowing Library',
+    version: '1',
+    filebase: 'budgie-windowing-1.0',
+    libraries: ['-L${libdir}', '-lbudgie-windowing'],
+    requires: ['gtk+-3.0 >= 3.24.0']
```
EbonJaeger, line 33: Should it also require libxfce4windowing?

fossfreedom: That's line 9 isn't it?

EbonJaeger: Does that automatically get put in the .pc?

I guess, does this even matter? xD

EbonJaeger (approved):
LGTM, thanks!

Issue #704, opened 2025-04-22
[Bug] Caffeine Applet notifications do not appear


fossfreedom:
### Description

Checked the BDS settings for the applet to show notifications.

Nothing appeared when caffeine mode ended

### Budgie version

10.9.2

### Operating System

UB 25.04

### Steps to reproduce the issue

1. Via the panel settings, check the notification option
2. Enable caffeine mode
3. Wait until caffeine mode completes or is turned off 

### Actual result

No notification

### Expected result

Notification is expected

### Additional information

From journalctl I can see the following:

```
Apr 22 22:10:47 2504wayland budgie-daemon[2487]: settings.vala:376: Failed to send our Caffeine notification: Timeout was reached
Apr 22 22:11:47 2504wayland budgie-daemon[2487]: settings.vala:367: Failed to close previous notification: Timeout was reached
Apr 22 22:12:12 2504wayland budgie-daemon[2487]: settings.vala:376: Failed to send our Caffeine notification: Timeout was reached
```

Note - also tried under git main i.e. wayland - same observation.

PR #705, opened 2025-04-23
Handle caffeine boost notifications and brightness changes

fossfreedom:
## Description
Resolves https://github.com/BuddiesOfBudgie/budgie-desktop/issues/704

Also ensures that if we have caffeine boost with max brightness, that this doesn't inadvertently invoke the brightness OSD to display

### Submitter Checklist


JoshStrobl (approved):
LGTM

PR #706, opened 2025-04-26
Only invoke one action per popup - closes #700

### LINKED ISSUE #700: [Bug] Notifications should only send one action on selecting an action button - it wrongly sends two
### Description

Raised from https://github.com/canonical/snapd-desktop-integration/issues/52#issuecomment-2809589984

"Ok, I found the problem: the notification manager of that desktop is sending wrong events. There are three events that a notification can send: "close" (when the notification is closed), "default" (when the user clicks over the notification, in the text), and a specific action name (when the user clicks on a button in a notification; for example, when pressing "yes" or "no" in the notification that asks whether to install or not the missing themes).

The problem is that, when the user clicks on a button, the notification manager sends the action for that button, followed by the "default" action; the correct behavior is to only send the button action, nothing else.

This explains the problems detected: since the "default" action installs the missing themes, clicking on "yes" sends two actions to install themes ("yes" and "default"), so the second one interferes with the first one, and that's why an error is shown, even when the theme is correctly installed. Also, pressing on "no" sends one action for "do nothing" ("no") and another action to install the themes ("default"), and that's why it installs it anyway.

In Gnome Shell and in Kde Plasma, the behavior is the expected one: only one action."

### Budgie version

10.9.x

### Operating System

Ubuntu Budgie 25.04

#!/usr/bin/env python3

import notify2
import dbus
import dbus.mainloop.glib
from gi.repository import GLib

loop = GLib.MainLoop()

dbus_mainloop = dbus.mainloop.glib.DBusGMainLoop()
notify2.init("test-notify", dbus_mainloop)

def cb(notification, action):
    print(f"Received action: {action}")

def closedcb(notification):
    print(f"Received CLOSE action")
    loop.quit()

n=notify2.Notification("A test notification", "Do click on the buttons, the close button, or the notification itself.", "")
n.add_action("yes", "Yes", cb)
n.add_action("no", "No", cb)
n.add_action("default", "Default", cb)
n.connect("closed", closedcb)
n.show()

loop.run()

### Actual result

shows two actions rather than one.

### Expected result

Should just show one action

### Additional information

_No response_

fossfreedom:
## Description
as per #700 - this ensures only one action is ever invoked per popup

Note - this PR also fixes an obvious spelling issue for a member function

### Submitter Checklist

### DIFF
--- a/src/daemon/notifications/dbus.vala
+++ b/src/daemon/notifications/dbus.vala
@@ -220,7 +220,7 @@
 		 * If replaces_id is not 0, the returned value is the same value as replaces_id.
 		 */
 		[DBus (name="Notify")]
-		public uint32 notication_received(
+		public uint32 notification_received(
 			string app_name,
 			uint32 replaces_id,
 			string app_icon,
--- a/src/daemon/notifications/popup.vala
+++ b/src/daemon/notifications/popup.vala
@@ -152,6 +152,7 @@ namespace Budgie.Notifications {
 		public bool did_interact { get; private set; default = false; }
 
 		private Body? body;
+		private bool actioned = false; /* only ActionInvoked once for each popup */
 
 		/**
 		 * Signal emitted when an action is clicked.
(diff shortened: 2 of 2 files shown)


JoshStrobl (approved):
LGTM :rocket:

PR #707, opened 2025-04-26
Always display the authentication user using a label if there is just one account - closes #624

### LINKED ISSUE #624: [Bug] Polkit dialog does not show authenticating user
### Description

The polkit dialog does not show which user is authenticating. This can cause user confusion, like in [this review](https://distrowatch.com/weekly.php?issue=20240122#solus):

> Finally, a security bug I mentioned in my last review remains and I was disappointed to see it lingers. Non-privileged accounts (regular user accounts) can perform administrative actions if they have the first user's password. This happens even when the regular user is not in the sudoer file and it allows guest users to install or remove software. This means if two users happen to have the same password (or similar ones and the user makes a typo) they can perform admin actions without being a privileged user. This seems like a design flaw and one that, while it probably can only be exploited in a few rare situations, makes me uncomfortable with the security settings on Solus. 

### Budgie version

10.9.2

### Operating System

Solus 4

1. Have a system with one admin user ('Test User' below).
1. Create an additional user without admin privileges ('Temp' below).
2. Do something that requires admin privileges. Eg: unlock the users page in the settings, restart a service without sudo.
3. Get a password prompt for the admin user, without mentioning that it's for the admin user.

### Actual result

afbeelding

### Expected result

Show the authenticating user when authenticating. Examples of this:

Polkit on terminal:

afbeelding

GNOME:

afbeelding

Plasma:

afbeelding

### Additional information

_No response_

fossfreedom:
## Description
See the headline

In summary - if just one authentication user is found then display the name via a centered label.  

Currently, if just one authentication user is found the name was hidden.

This ensure that even if a standard account is being used, the standard user will know they should be authenticating using the primary administrator.

### Submitter Checklist

### DIFF
--- a/src/dialogs/polkit/dialog.ui
+++ b/src/dialogs/polkit/dialog.ui
@@ -94,6 +94,35 @@
               </packing>
             </child>
             <child>
+              <object class="GtkLabel" id="label_username">
+                <property name="visible">False</property>
+                <property name="can-focus">False</property>
+                <property name="halign">start</property>
+                <property name="margin-top">6</property>
+                <property name="margin-bottom">8</property>
+                <property name="label" translatable="yes">Username:</property>
+                <property name="xalign">0</property>
+              </object>
+              <packing>
+                <property name="left-attach">1</property>
+                <property name="top-attach">2</property>
+              </packing>
+            </child>
+            <child>
+              <object class="GtkLabel" id="label_identity">
+                <property name="visible">False</property>
+                <property name="can-focus">False</property>
+                <property name="halign">start</property>
+                <property name="margin-top">6</property>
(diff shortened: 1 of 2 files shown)


fossfreedom:
picture taken from standard user showing just one administrator account.

image

EbonJaeger:
Hm. It looks really strange just floating there in the middle. Could we make it say something like "Authenticating as $name" and align it with Login Window and Password?

Or maybe it would look better if we added a pair of labels above the password so it looks like:

Username: dad
Password:  EntryBox

fossfreedom:
Snapshot_2025-04-30_20-58-01

EbonJaeger (approved):
LGTM, thanks!

PR #710, opened 2025-04-29
Make the screenshot dialog less aggressive to start when finding libxfce4windowing details.

fossfreedom:
## Description
We are expecting too much of libxfce4windowing to get the monitor and scale details on startup - the underlying wayland client needs more time after the desktop has initialised. This is intended to prevent crashing if the necessary details are not found before first usage. As a fallback, the autostart session is amended to restart if necessary if at anytime the dialog fails for any reason.

Taken together this makes the screenshot capability experience more robust

### Submitter Checklist


EbonJaeger (approved):
LGTM, thanks! Hopefully this means I don't have to kill or restart the screenshotter anymore the first time I want to use it.

PR #715, opened 2025-05-11
Port to libpeas 2

EbonJaeger:
## Description

This ports all the plugins and plugin handling code to libpeas version 2. We decided to take this route because of the upstream changes to GObject introspection breaking the Python plugin loader for libpeas-1. In order to keep support for Python applets, there was no choice but to migrate to libpeas-2.

At the beginning of the port, there was trouble getting the panel plugin manager to work in Vala. So, we decided to do it in C instead. It works now, but with minor modification, so does the Vala version in Raven, so it may not have been necessary. Oh, well. I like the new version better, anyway.

### Submitter Checklist


src/panel/manager.vala
```diff
@@ -612,12 +601,6 @@ namespace Budgie {
 				return;
 			}
 
-			/* Manual configuration from user met the expected migration path. Proceed as normal. */
-			if (panel_plugin_manager.migrate_load_requirements_met) {
-				message("Budgie Migration skipped due to user meeting migration requirements");
-				return;
-			}
-
 			message("Budgie Migration initiated");
```
fossfreedom, line 604: Can we remove all of the migration code now?  This was for a very old version of budgie-desktop

src/panel/plugin/meson.build
```diff
@@ -0,0 +1,61 @@
...
+    include_directories('.'),
+  ]
+)
+
+link_libpanelpluginmanager = declare_dependency(
+  link_with: libpanelpluginmanager,
+  include_directories: include_directories('.')
+)
+
+# Generate .gir
```
fossfreedom, line 31: were we intending to offer an api to-do plugin management?  Maybe clean this bit up?

src/panel/plugin/plugin-manager.c
```diff
@@ -0,0 +1,472 @@
...
+ */
+
+#include "plugin-manager.h"
+
+#include "budgie-config.h"
+#include "applet.h"
+#include "plugin.h"
+
+#include <gobject-introspection-1.0/girepository.h>
+#include <libpeas.h>
```
fossfreedom, line 19: Here we #include differently https://github.com/BuddiesOfBudgie/budgie-desktop/pull/715/files#diff-8099e7e0bcf95b8b2c50ce8111790e7632f31c816c07c5ffb6989243c6f385c0R16

src/panel/plugin/plugin-manager.vapi
```diff
@@ -0,0 +1,48 @@
+using Budgie;
```
fossfreedom, line 1: Any reason we aren't using gnome.generate_gir to create the vapi? as per https://github.com/BuddiesOfBudgie/budgie-desktop/blob/main/src/plugin/raven/meson.build#L60

src/panel/plugin/plugin-manager.c
```diff
@@ -0,0 +1,472 @@
...
+BudgieAppletInfo *budgie_panel_plugin_manager_load_applet_instance(BudgiePanelPluginManager *self, const gchar *uuid, GSettings *plugin_settings, gchar **name, GError **err) {
+	g_return_val_if_fail(BUDGIE_IS_PANEL_PLUGIN_MANAGER(self), NULL);
+	g_return_val_if_fail(uuid != NULL, NULL);
+	g_return_val_if_fail(name == NULL || *name == NULL, NULL);
+	g_return_val_if_fail(err == NULL || *err == NULL, NULL);
+
+	g_autofree gchar *path = NULL;
+	GSettings *settings = NULL;
+	g_autofree gchar *plugin_name = NULL;
+	PeasPluginInfo *info = NULL;
```
fossfreedom, line 359: The old API says that the caller owns the instance - so shouldn't we cleanup usage of info?

src/panel/plugin/plugin-manager.c
```diff
@@ -0,0 +1,472 @@
...
+	if G_UNLIKELY (!g_hash_table_contains(self->plugins, name)) {
+		g_set_error(err,
+		    BUDGIE_PANEL_PLUGIN_MANAGER_ERROR,
+		      BUDGIE_PANEL_PLUGIN_MANAGER_ERROR_NOT_LOADED,
+		    "Attempted to create a plugin that isn't loaded: %s", name);
+		return NULL;
+	}
+
+	path = create_applet_path(uuid);
+	settings = g_settings_new_with_path(BUDGIE_APPLET_SCHEMA, g_strdup(path));
```
fossfreedom, line 460: should we cleanup the usage of settings?

fossfreedom (changes requested):

src/panel/plugin/meson.build
EbonJaeger: Another thing I forgot to delete.

src/panel/plugin/plugin-manager.vapi
EbonJaeger: When I tried to, it failed to compile.

src/panel/plugin/plugin-manager.c
EbonJaeger: I have to double check the context, but I'm pretty sure the info we get is owned by Peas, so we dup it when we return it.

src/panel/plugin/plugin-manager.vapi
EbonJaeger: To clarify, when generating the GIR, it can't find the symbols from `src/config`.

Issue #716, opened 2025-05-14
[Bug] gdk_monitor_get_geometry assertion messages when running the panel


fossfreedom:
### Description

Spotted the following when running budgie-panel --replace &

```
budgie-panel:2941): Gdk-CRITICAL **: 20:55:57.892: gdk_monitor_get_geometry: assertion 'GDK_IS_MONITOR (monitor)' failed

(budgie-panel:2941): Gdk-CRITICAL **: 20:55:57.892: gdk_monitor_get_geometry: assertion 'GDK_IS_MONITOR (monitor)' failed

(budgie-panel:2941): Gdk-CRITICAL **: 20:55:57.893: gdk_monitor_get_geometry: assertion 'GDK_IS_MONITOR (monitor)' failed

(budgie-panel:2941): Gdk-CRITICAL **: 20:55:57.893: gdk_monitor_get_geometry: assertion 'GDK_IS_MONITOR (monitor)' failed

(budgie-panel:2941): Gdk-CRITICAL **: 20:55:57.894: gdk_monitor_get_geometry: assertion 'GDK_IS_MONITOR (monitor)' failed

(budgie-panel:2941): Gdk-CRITICAL **: 20:55:57.894: gdk_monitor_get_geometry: assertion 'GDK_IS_MONITOR (monitor)' failed

(budgie-panel:2941): Gdk-CRITICAL **: 20:55:57.894: gdk_monitor_get_geometry: assertion 'GDK_IS_MONITOR (monitor)' failed

```

### Budgie version

git master

### Operating System

UB 25.04

### Steps to reproduce the issue

See description

### Actual result

See description

### Expected result

best to see if we can clean this up

EbonJaeger:
They look to be coming from [here](https://github.com/BuddiesOfBudgie/budgie-desktop/blob/main/src/lib/toplevel.vala#L80). Struts aren't a thing that we use under Wayland, is it? Can we remove that whole function?

fossfreedom:
@EbonJaeger nice find - I'm not really famililar with this area so not sure what will happen.  Suppose coul delete and check.  Think this needs testing against multimonitor scenarios?

@JoshStrobl will the new display management stuff get involved in this area?

JoshStrobl:
I don't think we need this function at all, and instead use layer shell anchor setting (which we have already for panel position change).

PR #718, opened 2025-05-16
panel: Remove strut code

EbonJaeger:
## Description

We use GtkLayerShell on Wayland in place of struts, as struts aren't
really a thing on Wayland. This stops criticals from being printed when
the panel starts.

Fixes https://github.com/BuddiesOfBudgie/budgie-desktop/issues/716

### Submitter Checklist


JoshStrobl (approved):
LGTM, thanks!

PR #719, opened 2025-05-16
panel: Remove window intersection code

EbonJaeger:
## Description

We have no way of knowing a window's position under Wayland without
specific compositor integration, so don't even try. This should remove
more messages about `get_geometry()` not supported on Wayland.

### Submitter Checklist


JoshStrobl (approved):
LGTM, thanks!

PR #720, opened 2025-05-16
panel: Don't use geometry to detect monitor

EbonJaeger:
## Description

We can't get a window's geometry under Wayland, so instead, try to get
the "primary" monitor from the first Budgie panel we find, and check if
a window's monitor is the same.

It may be janky with workspaces, since we can't yet do
Window<->Workspace association with libxfce4windowing and labwc, but
otherwise, it *seems* to work. Kind of amazing, really.

### Submitter Checklist

### DIFF
--- a/src/panel/manager.vala
+++ b/src/panel/manager.vala
@@ -241,22 +241,43 @@ namespace Budgie {
 			check_windows();
 		}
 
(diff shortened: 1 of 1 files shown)


JoshStrobl (approved):
LGTM, thanks!

PR #724, opened 2025-05-27
Sync settings for both labwc and for apps that don't explictly take their cues from gnome gtk settings

fossfreedom:
## Description
This PR sets CSS settings overrides for cursor and theme settings in the users local config settings.ini file.  This is required for various apps such as Qt apps that need a helping hand to style correctly. 

Labwc also needs a similar helping hand setting the labwc-environment file with the settings vars - again helping labwc to style the interface.

### Submitter Checklist


EbonJaeger (approved):
LGTM, thanks!

PR #726, opened 2025-06-05
Full Screen shortcut

Kyuyrii:
This is just a small change so that the shortcut to activate full screen is already activated by default.

## Description

### Submitter Checklist


fossfreedom:
This is the wrong place. The bridge takes its defaults from the toggle-fullscreen dconf key. So this should either be done as an override in budgie-desktop, or left to distros to decide to override the gnome defaults.

PR #734, opened 2025-06-23
session: Add a start script to start the compositor

EbonJaeger:
## Description

This is adapted from Xfce4. The start script sets some environment variables, copies our configuration files to the correct user directory, and starts the Wayland compositor, in our case, labwc. They support running different compositors by passing it as an argument to the script, which is why they have the whole Opts thing. We may want that in the future too, so I left it in.

Note that if you're using Budgie Daemon V2, it will no longer autostart, presumably because we now set the config-dir for labwc. We could probably copy over the `/etc/xdg/labwc/autostart` file to the local dir, but I don't like that, because what if someone already has other entries in there, and I don't want to add yet another way to auto-start a program.

### Submitter Checklist

### DIFF
--- a/src/bridges/labwc/labwc_bridge.py
+++ b/src/bridges/labwc/labwc_bridge.py
@@ -59,7 +59,7 @@ class Bridge:
             mainloop=quit()
 
     def user_config(self, config_file="rc.xml"):
-        return os.path.join(GLib.get_user_config_dir(), "labwc", config_file)
+        return os.path.join(GLib.get_user_config_dir(), "budgie-desktop", "labwc", config_file)
 
     # writes the labwc rc.xml file back
     def write_config(self):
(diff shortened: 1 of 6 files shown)


fossfreedom:
Cheers Evan.

Ok - I've taken your route map here - works well.  I've scripted a session friendly autostart for BudgieDaemonV2 - it now starts before the panel/daemon.

This leads me to think about the purpose of the session folder here in this project.  Its probably a legacy thing where we could not fiddle with gnome-session.

Should we consolidate much of the session changes into budgie-session? - we being me - i'll put my hand up.  Think it will make things easier to understand.  At the moment we have session related stuff both here and in budgie-session and its kind of confusing where certain things happens.

EbonJaeger:
I think that this all should be moved into budgie-session. That would also help consolidate all of our session stuff in one place.

fossfreedom:
Evan - lets close this now as overtaken-by-events.  Cheers for the steer on this which was incorporated into the session-management-v3 branch.

PR #739, opened 2025-08-16
GNOME 49 GSD changes

fossfreedom:
## Description
Accommodate GNOME 49 changes.

Need to have a closer look at media keys
Brightness Control is now an issue since keyboard control doesn't seem to function anymore. .edit .... hopefully moving the helper deletion code and the policy stuff into budgie settings daemon is the dbus element needed 
Look at the wacom stuff https://gitlab.gnome.org/GNOME/gnome-settings-daemon/-/merge_requests/418/diffs

https://gitlab.gnome.org/GNOME/gnome-settings-daemon/-/commit/d23d7f6d3cf61b5854a25897fca0605a69164a01

### Submitter Checklist


src/daemon/bsd/plugins/meson.build
```diff
@@ -0,0 +1,185 @@
...
+if not enable_cups
+  disabled_plugins += ['print-notifications']
+endif
+
+# Specify futher required units, 'before' or 'after' may be specified if ordering is needed
+plugin_gate_units = {
+    'xsettings': [
+      # Both after/before. after for stopping reliably, before for synchronisation
+      ['gnome-session-x11-services.target', 'after'],
+      ['gnome-session-x11-services-ready.target', 'before'],
```
JoshStrobl, line 60: Note to self: Need to double check if these have been nuked in GNOME Session or Fedora.

JoshStrobl (changes requested):
Are there any files of note that I should take a look at otherwise? Otherwise it'll be a bunch of GSD code I'll need to go through that probably hasn't changed at all :D

src/daemon/bsd/gnome-settings-daemon.doap
JoshStrobl, line 1: Can this be ripped out as well?

src/daemon/bsd/plugins/meson.build
JoshStrobl: These still exist in F43 gnome-session at least as of 49.rc

src/daemon/bsd/gnome-settings-daemon.doap
fossfreedom: yep

src/daemon/bsd/plugins/meson.build
```diff
@@ -0,0 +1,185 @@
...
+if not enable_wwan
+    disabled_plugins += ['wwan']
+endif
+
+if not enable_colord
+  disabled_plugins += ['color']
+endif
+
+if not enable_cups
+  disabled_plugins += ['print-notifications']
```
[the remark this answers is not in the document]
fossfreedom: from the rest of the meson.build "disabled_plugins" is used in various places. Risk/Reward tilts for the former rather than the latter.

PR #743, opened 2025-09-13
Port libpeas-2 support to Budgie 10.9.x

JoshStrobl:
## Description

This PR ports over the work done by @EbonJaeger on moving Budgie Desktop (10.10) over to libpeas-2 and newer g_irepository APIs. This is a partial port as some changes such as typelib 2.0 for the ABI bump isn't relevant to Budgie 10.9.x.

This support should, from a distribution's perspective, reduce the amount of reverse dependencies on libpeas1 as we'll have moved to 2.0. It should also reduce any potential headaches related to the whole "libpeas1 / gobject-introspection-1.0 / pygobject3" story.

I took the opportunity to port over work from @fossfreedom on vapi cleanup, this was particularly useful since we had a lot of redundant references including undesired ones pointing to vapi files for libpeas1. Some elements were restored for X11 compat (particularly with Raven calling GdkX11 APIs in Budgie 10.9.x).


EbonJaeger (approved):
LGTM! Though I might be biased since I was the one that first wrote this code xD

JoshStrobl:
@EbonJaeger Yea I'd hope you'd approve of your own code. At least in the general sense, even if we all at some point look back at our own code with dismay.

fossfreedom:
hmm. Without the plugin version bump wouldn't a user who is using an existing peas1 plugin on the panel potentially competing with builtin peas2 plugins? Potential panel crash scenarios? A user upgrading will be left wondering what has happened and how to recover.

JoshStrobl:
> Without the plugin version bump wouldn't a user who is using an existing peas1 plugin on the panel potentially competing with builtin peas2 plugins?

Hmm possibly.  I haven't tested a scenario of having libpeas1 and libpeas2 plugins installed at the same time (only tested with default applets). I am not opposed to a bump to 2.0 and making sure our plugins are all bumped to 2.0 if necessary, we would just want then make sure 10.10 has 3.0 instead of 2.0 for the reason of the original bump (indicator of Wayland compatibility).

JoshStrobl:
Will ABI bump so we can get this in, let's remember to re-re-bump 10.10 main to 3.0

fossfreedom:
> Will ABI bump so we can get this in, let's remember to re-re-bump 10.10 main to 3.0

Hmm.... major brain wobble. In main we currently don't bother with any plugin/raven-widget version check - do we?

Think we just assume plugins have been built with peas-2.

JoshStrobl:
@fossfreedom That's correct. It's basically "oh we found it in the gi repo, alright load it in no questions asked". But since one has to compile against our headers / pkgconfig and those use libpeas-2, should be OK.

Bumped Raven and pkgconfigs. Tested on laptop and both panel and raven widgets load.

PR #748, opened 2025-10-11
Enabling thumbnail in Labwc window switcher

Kyuyrii:
This is just a small tweak to enable thumbnails in the Labwc window switcher. This feature was added in the latest version, 0.9.2.

I tested it on Ubuntu Budgie 25.04 with Budgie 10.9.2, and it worked without any issues.


fossfreedom:
https://github.com/user-attachments/assets/82cfb49a-e890-4f46-a5e9-48b9f3dbcfa7

PR #753, opened 2025-10-17
Added use of QT Oxygen style

Kyuyrii:
## Description
I made this small modification to improve the appearance of QT apps, can you see if it is done correctly?

### DIFF
--- a/src/session/startbudgielabwc.in
+++ b/src/session/startbudgielabwc.in
@@ -65,6 +65,12 @@ if [ -z "${QT_QPA_PLATFORMTHEME}" ]; then
     export QT_QPA_PLATFORMTHEME
 fi
 
+# Improves the appearance of QT apps when using the QT Oxygen style
+if [ -z "${QT_STYLE_OVERRIDE}" ]; then
+    QT_STYLE_OVERRIDE="Oxygen"
+    export QT_STYLE_OVERRIDE
+fi
+
 # In many OS the wayland backend for clutter does not work, so
 # use the safe "gdk" backend instead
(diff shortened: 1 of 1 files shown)


JoshStrobl:
I appreciate you taking the time to open this, however after consultation with some Qt developers in the space, we have reached the conclusion that setting this would break more than it would help. Therefore we will not be shipping this change.

There are mechanisms that exist already for end users to override environment variables that wish to use specific Qt styling, so I would suggest you explore those and use it to apply this to your own session.

Kyuyrii:
> There are mechanisms that exist already for end users to override environment variables that wish to use specific Qt styling, so I would suggest you explore those and use it to apply this to your own session.

In that case, which one are you talking about?

The lxqt-qtplugin isn't supported by Snap/Flatpak apps; I assume the same applies to qt6ct.

The only way I know to choose a QT style is by editing the /etc/environment file to add, for example, QT_STYLE_OVERRIDE=Oxygen.

The default style used is Fusion, which many users may not even know is a QT style or that it can be changed.

I only learned about QT_STYLE_OVERRIDE because I was messing around with PyQt6.

And I've always been bothered by the huge difference in appearance of QT apps in KDE and GTK DEs, since KDE uses Breeze, while other DEs use Fusion.

Furthermore, the Oxygen QT style behaves more closely to the Windows QT style.

I don't have access to Oxygen here, so I'll show you a QT Fusion style problem that doesn't happen in QT Windows and Oxygen style.

Captura de tela de 2025-10-18 14-48-33
Captura de tela de 2025-10-18 14-49-00

PR #754, opened 2025-10-18
Session management v3

fossfreedom:
## Description
Leverages Evans session work in the session-management-v2 branch (i.e. replaces that PR)
- Adds a revised method to open labwc with our own config files rather than the default labwc .config folder
- Now waits for labwc to fully start before initiating budgie-session
- Incorporates GNOME 49 revisions - our media keys schema and adjusts the bridge to use those schema keys
- Adds Josh's budgie-daemon-v2 display startup as a session item
- Adds QT6 styling via the environment variable QT_QPA_PLATFORMTHEME=gtk3

### dependencies
Requires:
- git main build of budgie-session + the PR for the helper compositor check
- git main build of budgie-daemon-v2
- git main build of budgie-control-center

### notes
- the heart of this is the startbudgielabwc bash script launched by the /usr/share/wayland-session/budgie-desktop.desktop
- best remove your ~/.config/labwc and ~/.config/budgie-desktop/labwc folders to force a pull-down of the default config files
~~If we want QT Oxygen styling out of the box we should also merge in @Kyuyrii PR into this branch - Oxygen styling is better for flatpak / snap based apps.~~
- @JoshStrobl I note budgie-daemon-v2 does not provide an autostart desktop file.  In src/session I've added a desktop file here to launch budgie-daemon-v2.  If the autostart desktop file is added to the v2 project then it should be removed here in this project/branch
- @JoshStrobl budgie-daemonV2 is the only process that somehow survives a logout - problem is that if you relogin again then two budgie-daemonV2 processes are available causing GUI lockups
- tested with labwc 0.9.2 and wlroots 0.19
~~Observation 1: X11 based apps do not appear to launch as an autostart - for example nemo-desktop.  This is the same observation with wayfire. It appears with the autostart stuff occurs, DISPLAY=:0 is not available in the environment. XWAYLAND should do this.  running nemo-desktop from a terminal post login launches correctly via xwayland~~ fixed with the reuse of labwc -S
~~Observation 2: Firefox on ubuntu 25.10 doesnt launch via the icon-task-list.  In the logs in mentions the missing display number.  Odd.~~ fixed with the reuse of labwc -S  
- Observation 3: on very first login, the desktop gtk theme is not applied - it is when next logged in.  Not sure why at the moment.  Needs investigating.
- GNOME-Boxes (QEMU install) with the above now successfully works without any black screens
- Virtualbox 7.2.4 with/without 3D accel either does not start the desktop or the desktop crashes few seconds after login

### Submitter Checklist

### DIFF
--- a/src/session/budgie-desktop.in
+++ b/src/session/budgie-desktop.in
@@ -13,7 +13,7 @@ if [ -z $XDG_CURRENT_DESKTOP ]; then
   export XDG_CURRENT_DESKTOP
 fi
 
-@bindir@/org.buddiesofbudgie.BudgieDaemonV2 &
+@bindir@/org.buddiesofbudgie.Services &
 sleep 2
 
 # Wait until the server starts before launching the session


JoshStrobl (changes requested):
Seems good to me but would like an explanation on the rc.xml change.

JoshStrobl (approved):

PR #755, opened 2025-10-30
Check if GTK theme file is exists

City-busz:
## Description
This ensures that only those themes are added to the list, which actually provide a theme for GTK3. E.g. themes are only for Openbox are not listed.

### Submitter Checklist


fossfreedom (approved):
Seems reasonable. We don't need to worry about gtk2 stuff. Nor do we need to  worry about pure gtk4 themes that just have a gresource file.

PR #759, opened 2025-10-31
Set colors for focused item in the main menu

City-busz:
## Description
This makes the keyboard navigation more usable as the actually focused item is highlighted.

### Submitter Checklist


fossfreedom (approved):
Had a good rummage around this looking at colours, focus issues etc.  Didn't spot any adverse effects.

Looks good to merge.  Much appreciated!

PR #761, opened 2025-11-01
Show screenshot tool only in Budgie session

City-busz:
## Description
It relies on the `org.buddiesofbudgie.BudgieScreenshotControl` D-Bus interface, which is specific for Budgie.

### Submitter Checklist


fossfreedom (approved):
LGTM!

PR #772, opened 2025-11-19
Fix critical crashes and robustness issues across Budgie Desktop

### LINKED ISSUE #751: [Bug] Crash if icon theme contains broken symlinks
### Description

If you have a GTK icon theme that happens to contain a broken symlink, Budgie crashes and will not recover upon reboot. This can happen if one theme is a variation on another (such as dark-and-light versions) and the user removes one of them.

The only solution is to log into a non-GUI terminal session (or possibly use another desktop, if you have one) and change the theme (or fix the symlinks) to one that isn't broken.

```sh
gsettings set org.gnome.desktop.interface icon-theme 'hicolor'
```

I feel strongly that the presence or absence of icons should not crash the desktop.

### Budgie version

10.9.2

### Operating System

Solus

1. replace an icon used by Budgie with a broken symlink

### Actual result

crash, desktop unusable, even across reboots

### Expected result

desktop falls back to other icon or shows a generic placeholder

### Additional information

_No response_

Lib-LOCALE:
This PR fixes multiple critical crashes and robustness issues discovered through code analysis:

Added null check before load_icon() in IconTasklistApplet.vala
Prevents crash when icon theme contains broken symlinks
Additional Critical Fixes
Background daemon crashes: Fixed NULL pointer dereferences in background.vala
Notification crashes: Added validation for hint lookups in dbus.vala
Bluetooth crash: Fixed triple NULL dereference in BluetoothObexManager.vala
Resource leak: Added destructor to cleanup DBus subscription in osdkeys.vala
Array bounds: Added bounds checking in multiple applets
URI parsing: Added validation for malformed URIs

Closes #751

### DIFF
--- a/src/panel/plugin/plugin-manager.c
+++ b/src/panel/plugin/plugin-manager.c
@@ -378,7 +378,10 @@ BudgieAppletInfo *budgie_panel_plugin_manager_load_applet_instance(BudgiePanelPl
 	}
 
 	plugin_name = g_settings_get_string(settings, BUDGIE_APPLET_KEY_NAME);
-	info = g_hash_table_lookup(self->plugins, plugin_name);
+
+	if (g_hash_table_contains(self->plugins, plugin_name)) {
+		info = g_object_ref(g_hash_table_lookup(self->plugins, plugin_name));
+	}
 
 	// Check if the plugin has been loaded
 	if (!PEAS_IS_PLUGIN_INFO(info)) {


Lib-LOCALE:
Hi @EbonJaeger! Yes, this patch was created with the assistance of Claude (Anthropic's AI assistant) via claude.ai. I used Claude to help analyze the codebase, identify the crash issues, and generate the fixes. I then reviewed and tested the changes before submitting. I'm happy to make any adjustments you suggest!

Lib-LOCALE:
@EbonJaeger I've addressed your feedback in commit 0bc3a3d2:

✅ Changed `warning()` to `critical()` for the null Bluetooth Obex session check
✅ Added window names to the class ID warnings in ButtonPopover for better debuggability

The warnings now display: `"Window '%s' has no class IDs"` with the actual window name.

Thanks for the review!

PR #776, opened 2025-11-29
Add environment variables to help running budgie in virtualbox and also help electron based apps think they are still running in GNOME

fossfreedom:
## Description
Helper environment variables as per the title

### Submitter Checklist

### DIFF
--- a/src/session/startbudgielabwc.in
+++ b/src/session/startbudgielabwc.in
@@ -76,6 +76,14 @@ if [ -z "${SDL_VIDEODRIVER}" ]; then
     export SDL_VIDEODRIVER
 fi
 
+# Add a wlroots helper to allow running in VirtualBox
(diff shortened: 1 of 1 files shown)


JoshStrobl:
@fossfreedom which Electron apps was this tested against? At least the behavior for Signal isn't changed, it still shows the prompt about secret storage.

fossfreedom:
Josh - these apps https://discourse.ubuntubudgie.org/t/many-appindicator-icons-missing-under-25-10/8476

JoshStrobl:
Tested and no differences in my session (regression or otherwise). Of course, I'm using stock System Tray where my tray icons for apps like Element, Proton Pass, Ferdium, etc. have always worked, but no reason for it to be broken elsewhere :)

PR #778, opened 2025-12-15
fix(panel): crash on direct position flip (horizontal or vertical)

JoshStrobl:
## Description
There is something gnarly happening either internal;y in GTK show for the toplevels, or with the compositor, that triggers the panel to crash when we attempt to perform a direct position change (e.g. bottom to top).

Fully reconstructing the panel (closing it, removing it from panels hashtable, loading it again) seems to fix it, with the terrible caveat that it breaks conflicting panel position logic.

We'll want to fix duplicate panel position logic in Budgie Desktop Settings later.

### Submitter Checklist

### DIFF
--- a/src/panel/manager.vala
+++ b/src/panel/manager.vala
@@ -1112,30 +1112,26 @@ namespace Budgie {
 				return;
 			}
 
-			// This is horrible and I know it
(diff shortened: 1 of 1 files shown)


EbonJaeger (approved):
I don't pretend to understand this dark magic, and I'm sorry that you do.

PR #782, opened 2025-12-20
chore: rename desktop id to bdGTK

JoshStrobl:
## Description
This rename (we were originally Budgie:GNOME, then just Budgie, and now bdGTK) is to better enable differentiation between Budgie 10.10 (and possibly 10.9.x with a backport) from future Budgie 11 for various projects that utilize exact or substring matches for logic for Budgie Desktop.

This will be useful in places such as xdg-utils where we can do Budgie pre-11 (so leveraging GTK) logic that can be more easily decoupled / disassociated from Budgie 11 (Qt6).

### Submitter Checklist


JoshStrobl:
CI needs to fixed because we killed budgie-screensaver in Fedora, should be OK otherwise though!

gmbr3:
While this is one option - which is okay by me

The other option is to follow a more KDE way of checking (especially for xdg-utils) which keeps the same XDG_CURRENT_DESKTOP but exposes version information to the command line via KDE_SESSION_VERSION (4,5,6)

Conan-Kudo:
I think we'll do that with Budgie 11, but it's more uplift than we really want to invest in for Budgie 10 series.

EbonJaeger:
Well, where would that be done? If it's a really simple change, I would actually like that better than this.

fossfreedom:
> Well, where would that be done? If it's a really simple change, I would actually like that better than this.

I believe this means introducing a new variable, e.g., BUDGIE_SESSION_VERSION (maybe set to 10 for current Budgie 10.x series, 11 for the next major release, etc.) - aka KDE_SESSION_VERSION
Tools (especially xdg-utils scripts, or any Budgie-specific integrations) could then check the version and adapt behavior without needing to change the main XDG_CURRENT_DESKTOP string, avoiding breakage for apps that expect "Budgie".

^^^^
thats AI gemini speaking BTW

fossfreedom:
Having a quick look at the budgie-session stuff - the session checks are very explicit - it just looks for the given session name from the environment - so might need a rethink to say "if budgie or bdGTK" throughout. Not a 5 minute change. It should cope if the XDG_CURRENT_DESKTOP is "Budgie:bdGTK:wlroots" etc.  Need to double check that.

PR #784, opened 2025-12-22
Remember display changes from wdisplays

fossfreedom:
## Description
**This isn't meant for merging.**

This is just a patch to enable displays to remember what they were, position and scaling.  It will be reapplied when you login.

Requires wlr-randr and wdisplays - it hard-codes a daemon exe in /usr/libexec/budgie-desktop/wlr-display-manager launched from budgie-daemon.

Patch will be dropped from Debian/Ubuntu once budgie-desktop-services/budgie-display-configurator is accepted in Debian/Ubuntu and we (buddies) consider it is ready.

### Submitter Checklist


Conan-Kudo:
This should be obsolete with the finalized 10.10.0 code leveraging budgie-desktop-services.

PR #786, opened 2025-12-22
labwc: Use new snapping range config

EbonJaeger:
## Description
labwc 0.9.3 changed the snapping range syntax in the `rc.xml` config.

### Submitter Checklist


fossfreedom:
nice catch - tested this on v0.9.3 that has just landed in 26.04

fossfreedom (approved):

PR #788, opened 2025-12-24
[10.9] feat(session): Add BUDGIE_SESSION_VERSION for external tool handling

gmbr3:
## Description
An alternative solution for xdg-utils

### Submitter Checklist


fossfreedom (approved):
This changes budgie to use the well understood KDE version paradigm.  So IMHO easier for the wider ecosystem to implement per budgie version controls.

PR #790, opened 2025-12-24
Connect to the top-level focus-out event to to close the popover

fossfreedom:
## Description
Ensure the popover closes if clicking on any other window.

Thoughts on this tiny change?

### Submitter Checklist

### DIFF
--- a/src/plugin/panel/popover-manager.c
+++ b/src/plugin/panel/popover-manager.c
@@ -30,6 +30,7 @@ struct _BudgiePopoverManagerPrivate {
 G_DEFINE_TYPE_WITH_PRIVATE(BudgiePopoverManager, budgie_popover_manager, G_TYPE_OBJECT)
 
 static void budgie_popover_manager_widget_died(BudgiePopoverManager* self, GtkWidget* child);
+static gboolean on_focus_out(GtkWidget *widget, GdkEvent *event, GtkPopover *popover);
 
 /**
  * budgie_popover_manager_new:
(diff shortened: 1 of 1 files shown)


EbonJaeger (changes requested):

JoshStrobl:
LGTM and Evan's PR item addressed but will let him do the honors of approving as opposed to me railroading it into main :D

src/plugin/panel/popover-manager.c
```diff
@@ -143,6 +146,11 @@ void budgie_popover_manager_show_popover(BudgiePopoverManager* self, GtkWidget*
 	gtk_layer_set_keyboard_mode(popover_win, GTK_LAYER_SHELL_KEYBOARD_MODE_ON_DEMAND);
 }
 
+static gboolean on_focus_out(GtkWidget *widget, GdkEvent *event, GtkPopover *popover) {
+	gtk_widget_hide(GTK_WIDGET(popover));
+	return GDK_EVENT_PROPAGATE;
+}
```
EbonJaeger: Thinking more on it, I think we should be extra safe and make sure that `popover` is actually valid before trying to use it.

```suggestion
static gboolean on_focus_out(GtkWidget *widget, GdkEvent *event, GtkPopover *popover) {
	g_return_val_if_fail(popover != NULL, GDK_EVENT_PROPAGATE);

	gtk_widget_hide(GTK_WIDGET(popover));
	return GDK_EVENT_PROPAGATE;
}
```

EbonJaeger (changes requested):

src/plugin/panel/popover-manager.c
fossfreedom: agreed.  Tested and confirmed ok.

Issue #472, opened 2023-10-09
Implement Wayland support in legacy Tasklist applet


JoshStrobl:
This is a meta-tasklist for implementation of Wayland support in Legacy Tasklist applet, such as via use of [libxfce4windowing](https://gitlab.xfce.org/xfce/libxfce4windowing). Strict usage of Wnck is not viable as that is limited to X11. libxfce4windowing is designed to support both X11 and Wayland.

EbonJaeger:
Initial work on this can be found in the [simple-tasklist](https://github.com/BuddiesOfBudgie/budgie-desktop/tree/feature/simple_tasklist) branch. It even pretty much works, I think. It should most likely be ported to using the Windowing module in the icon-tasklist branch.

PR #793, opened 2026-01-05
feat(applets): implement tasklist replacement with Wayland support

### LINKED ISSUE #95: Improving the legacy Tasklist
This is a meta task to highlight actionable items that would improve the current legacy Tasklist. This is wholely inspired by recent development efforts related to the IconTasklist (#25). To ease discussion, I will quote the closing comment from that discussion below:

> I am going to be closing this issue filed as since the posting and the further work we have done on the IconTasklist, it has become abundantly clear that we cannot make the necessary changes and architectural improvements to the applet while facing repeated regressions due to the complexity of all the options provided.
>
> Specifically, almost all the regressions in IconTasklist are the result of needing to handle both grouping and non-grouping scenarios. Grouping is much easier to handle thanks to our robust application tracking system, with naming being consistent and most edge cases being worked out. With non-grouping, we need to track both that as well as the xid. This introduces problems for some applications like LibreOffice and Steam, which are otherwise well handled in grouping. We also have to use logic that ensures we re-parent secondary applications from their IconButton to the "main" one, handled pinned cases, etc.
>
> So Budgie 10.6.x will have the non-grouping support, however it will be removed in 10.7. We will introduce a new legacy Tasklist that lacks grouping, has labels, and likely no (or a simpler) popover for some operations like close, move to workspace, etc. This will be in a new task.

**So what does this mean for the Tasklist?**

Instead of effectively merging the Tasklist, which is a non-grouping and simplistic applet leveraging Wnck.Tasklist, into the IconTasklist applet, we will take this opportunity to rewrite the Tasklist to meet the needs of user desiring a non-grouping list of applications. Due to the fact the IconTasklist will not need non-grouping logic any further, this will greatly improve its implementation from an architectural point of view.

The new Tasklist will leverage our application tracking system instead of Wnck, which was effectively a goal with Budgie 11 anyways (elimination of Wnck since it won't be usable on Wayland), to provide a simple list of labeled and non-grouped applications. We may also be able to re-use our icon scaling code from the IconButton in the individually labeled buttons in the Tasklist.

I am unsure it is desired from an architectual point of view to have "pinning", as we would get into the same situation we are in with IconTasklist with re-parenting non-primary instances onto the original button when the original instance of the app is closed. Discussion around that is welcome.

JoshStrobl:
## Description
This pull request introduces a new Tasklist replacement. Originally, this was going to be referred to as the "Simple Tasklist", which takes a hybrid approach between the IconTasklist and Tasklist, however given our move to Wayland for Budgie 10.10 and Wnck being non-functional under Wayland (instead we've been swapping Wnck for libxfce4windowing), "Simple Tasklist" is simply now _the_ Tasklist.

This work is the culmination of a ton of work between @guillotjulien and @EbonJaeger (over the years, so massive thanks to them!), with some minor changes from my side (panel section width logic to ensure applets reach a max width / height state, and some UX tweaks).

Resolves #95 and #472 

### Submitter Checklist


EbonJaeger (approved):
LGTM, thanks!

PR #797, opened 2026-01-10
chore(bridge): change how we generate environment file so we aren't enforcing a given layout default

JoshStrobl:
## Description
This pull request changes how we generate the environment file in the labwc bridge.

1. We no longer actually ship an environment file for labwc. We generate it on the fly and fully regenerate it when various settings change (keyboard layout / order, cursor props).
2. We now avoid having U.S. International or UK layout being set via an environment file and relying on an override.
3. We no longer use a subprocess with sed. Instead, we fully regen the file with our properties.

### Submitter Checklist

### DIFF
--- a/src/bridges/labwc/labwc_bridge.py
+++ b/src/bridges/labwc/labwc_bridge.py
@@ -294,7 +294,7 @@ class Bridge:
         if layout == "":
             layout = "us" # default to at least a known keyboard layout
 
-        lines.append("XKB_DEFAULT_LAYOUT=" + layout + "\n")
+        lines.append(f"XKB_DEFAULT_LAYOUT={layout}\n")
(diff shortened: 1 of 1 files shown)


fossfreedom:
> 2\. We now avoid having U.S. International or UK layout being set via an environment file and relying on an override.

What override? a gsettings override or something else?

JoshStrobl:
I'm thinking along the lines of the live environment for Fedora Budgie and how gsd internally may be trying to determine default keyboard layout with locale1. I'm a bit paranoid about the pre-existing default value that labwc would use resulting in side effects between Anaconda, locale1, gsd, Budgie where the layout doesn't correctly propagate.

Conan-Kudo:
This was happening while I was testing Budgie 10.10 in Fedora, where despite my system settings and g-s-d settings, I wasn't getting my correct keyboard so that I can type properly. This PR aims to fix that.

PR #810, opened 2026-01-18
Remove GTK theme blacklist

City-busz:
## Description
Since commit 673231e658216ce72677985377c17ecf1ea839f3 the existence of the GTK3 CSS is explicitly checked, there is no need to add those themes to the blacklist which does not provide a GTK3 theme variant. The only themes that actually provide a GTK3 variant is Adwaita and Breeze, but these GTK themes work without issues, only the corresponding icon themes are problematic.

Those themes are already filtered out since they don't provide a GTK3 variant:
- Clearlooks: gtk2+openbox theme
- Crux: xfwm4 theme
- Default: gtk3 keybindings
- Emacs: gtk3 keybindings
- Industrial: gtk2 theme
- Kvantum: gtk2 theme
- Mist: gtk2 theme
- Murrina: gtk2 theme
- Raleigh: gtk3+metacity theme
- Redmond: xfwm4 theme
- ThinIce: gtk2 theme

### Submitter Checklist

### DIFF
--- a/src/panel/settings/themes.vala
+++ b/src/panel/settings/themes.vala
@@ -44,22 +44,6 @@ class ThemeInfo : GLib.Object {
 public class ThemeScanner : GLib.Object {
 	string[]? xdg_paths = null;
 
-	string[] gtk_theme_blacklist = {
-		"Adwaita",
-		"Breeze",
-		"Clearlooks",
-		"Crux",
-		"Default",
-		"Emacs",
-		"Industrial",
-		"Kv",
-		"Mist",
-		"Murrina",
-		"Raleigh",
-		"Redmond",
-		"ThinIce"
-	};
-
 	string[] icon_theme_blacklist = {
 		"Adwaita",
 		"HighContrast",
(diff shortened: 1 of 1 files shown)


JoshStrobl:
> but these GTK themes work without issues

I disagree with this assertion. Themes like Adwaita and Breeze are not designed with Budgie Desktop support (e.g. our classes in mind) and therefore turning off Built-in theming can result in breakages. I have no desire to remove them from the blacklist.

The themes which expose only GTK2, sure those can be removed.

City-busz:
> turning off Built-in theming can result in breakages

What kind of breakages you mean? I can't see these if I set Adwaita or Breeze. But there are a lot of GTK themes that are not designed with Budgie Desktop support. Maybe it would better to force the built-in panel theme if the user selects a GTK theme that is known to be broken with Budgie. Or just set some fallback CSS rules in this case.

JoshStrobl:
> What kind of breakages you mean?

You can easily see it by settings Adwaita via gsettings and turning off built-in theme. IconTasklist padding on IconButtons is off and clips icons, spacing is off for tray items, Raven headers are missing any form of contrast, etc.

I am not removing Adwaita from the blocklist. I am fine with removing the clearly GTK2 ones from being selected, but the rest stays.

City-busz:
> You can easily see it by settings Adwaita via gsettings and turning off built-in theme. IconTasklist padding on IconButtons is off and clips icons, spacing is off for tray items, Raven headers are missing any form of contrast, etc.

I couldn't see these:
budgie-adwaita

JoshStrobl:
If you don't consider that bad, I don't know what to tell you, but it's not happening.

City-busz:
Yes, it's the plain default style of GTK. Maybe it's not as fancy as the built-in theme, but it's usable. Users always have the option to turn on the build-in theme if they want.

JoshStrobl:
I really don't know how many other ways there are to tell you that I'm not removing it or Breeze. If you wish to remove the others with an updated PR, then go for it, but those two are absolutely staying.

City-busz:
Breeze GTK theme looks also good:
budgie-breeze

JoshStrobl:
While I appreciate the pull request, this introduces a clear regression and will not be included. Therefore, closing.

City-busz:
It's not a regression. It's an improvement.

PR #813, opened 2026-01-19
Autostart gsd-keyboard

City-busz:
## Description
It's needed to set the initial input sources based on the keyboard layout configured in systemd-localed.

Without this, the labwc bridge does not get the configured layout, and falls back to 'us'.
This is also used by the keyboard-layout applet.

### Submitter Checklist

### DIFF
--- a/src/session/meson.build
+++ b/src/session/meson.build
@@ -11,6 +11,7 @@ budgie_session_324_components = [
     'org.buddiesofbudgie.SettingsDaemon.Datetime',
     'org.buddiesofbudgie.SettingsDaemon.DiskUtilityNotify',
     'org.buddiesofbudgie.SettingsDaemon.Housekeeping',
+    'org.buddiesofbudgie.SettingsDaemon.Keyboard',
     'org.buddiesofbudgie.SettingsDaemon.Power',
     'org.buddiesofbudgie.SettingsDaemon.PrintNotifications',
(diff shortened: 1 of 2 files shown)


fossfreedom:
hmm - check in journald - doesn't keyboard start spewing out logs of various sorts - mainly around dbus issues I think?

City-busz:
No, I couldn't see such errors in the journal.

The source code of gsd-keyboard is pretty simple: it just connect to systemd-localed, read the settings, and save it into GSettings if not already configured:
https://gitlab.gnome.org/GNOME/gnome-settings-daemon/-/blob/9add93c492bae5400138ee872bd64009dbf48350/plugins/keyboard/gsd-keyboard-manager.c

fossfreedom:
Fair enough - let me have a play with this before merging.  Need to refresh my mind why I removed it in the first place.... maybe I just saw "keyboard" and said "don't need that"!

killajoe:
Just have the issue with EndeavourOS/Archlinux that keyboard layout is unset.
And force gsd-keyboard to autostart is solving the issue. 

Will this get merged anytime soon?  
Only to specify, at least on Archbased it will leave keyboard layout unset as it is now shipped with archlinux packages, you can set layout over settings indeed. But looks bad, as of a Desktop should apply what it has in system locale.

fossfreedom:
Many thanks for offering this PR. The now merged bridge changes should resolve locale on login issues.

PR #814, opened 2026-01-19
Add option to disable sync with GtkSettings file

City-busz:
## Description
This allows the users to disable writing the cursor and theme settings in the user's local config settings.ini file. Saving into that file is not always desirable, because it's applied also for other sessions.

### Submitter Checklist


City-busz:
@fossfreedom, why did you revert this in commit https://github.com/BuddiesOfBudgie/budgie-desktop/commit/0abc4ce7112388eb09b004e7de94b77e90597d50?

PR #816, opened 2026-01-19
Save window buttons layout directly to gnome wm schema

City-busz:
## Description
This eliminates the need to keep the values in sync with the setting stored in Budgie GSettings schema, and also allows users to set custom button layouts for their needs.

### Submitter Checklist

### DIFF
--- a/src/daemon/settings.vala
+++ b/src/daemon/settings.vala
@@ -114,9 +114,6 @@ namespace Budgie {
 			gnome_power_settings.changed["sleep-inactive-ac-timeout"].connect(this.update_ac_timeout);
 			gnome_power_settings.changed["sleep-inactive-battery-timeout"].connect(this.update_battery_timeout);
 
-			wm_settings.changed.connect(this.on_wm_settings_changed);
-			this.on_wm_settings_changed("button-style");
-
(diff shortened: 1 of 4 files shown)


JoshStrobl (changes requested):
While I understand the premise of the change, I do not agree with the implementation. Budgie 10.10 is in maintenance mode and I want to minimize schema changes, especially ones that have been around for years.

If you want to add a `custom` button-style enum and expose an option when that is selected in Budgie Desktop Settings to enter your own, I'm open to that, but the current solution otherwise remains.

City-busz:
This change won't break anything since the setting is already in sync with GNOME's settings, so I think it's safe to apply it.

JoshStrobl:
I'm informing you about the desired method for implementation should you wish for this to be included.

City-busz:
I don't think it would be useful to introduce more settings. It's already over-complicated. The purpose of this change is to simplify thinks.

JoshStrobl:
Closing as implementation doesn't align with requested. Feel free to open a new PR when it does.

City-busz:
This is the perfect implementation.

PR #818, opened 2026-01-20
Rework our libxfce4windowing wrapper to be resilient for repeated usage

fossfreedom:
## Description
This PR reworks the singleton class we have that wraps libxfce4windowing.

We now have moved this into a static library - this simplify the code greatly and removes alot of repeated code.

The key part here is adding more safe checks throughout and adding rate-smoothing to stop repeated calls that were occasionally causing mutex locks in libxfce4windowing itself.

Taken together this is more robust - running for a couple of days and screenshot and other stuff in budgie-daemon that was positioning things on the screen no longer crash/exit unceremoniously.

Please do try this on your setups

### Submitter Checklist

### DIFF
--- a/src/lib/waylandclient.vala
+++ b/src/lib/waylandclient.vala
@@ -1,0 +1,269 @@
+/*
+ * This file is part of budgie-desktop
+ *
+ * Copyright Budgie Desktop Developers
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2 of the License, or
+ * (at your option) any later version.
+ *
+ * The underlying WaylandClient does not appear to be fully thread-safe and either
+ * repeated calls very quickly, or calls within the same process where the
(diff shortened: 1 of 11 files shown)


src/lib/waylandclient.vala
```diff
@@ -0,0 +1,261 @@
...
+                Idle.add(() => {
+                    initialization_failed();
+                    return false;
+                });
+                return;
+            }
+
+            screen.monitors_changed.connect(on_monitors_changed_smoothed);
+            initialize_monitor_info();
+        }
```
JoshStrobl: Could we move the constructor above the getters?

fossfreedom:
I'll keep an eye out for this. But good that the stability is much better.  key for me was screenshot not dying on me at pretty random times.

JoshStrobl:
Been using it as well and it's been a blessing just being able to not have to launch budgie-screenshot-dialog in kitty after wake-from-suspend, so certainly approved from me too!

PR #819, opened 2026-01-22
Add support for different windows focus mode 

fossfreedom:
## Description
specifically allow sloppy focus as well as the standard click and the 10.9 mouse modes

### Submitter Checklist


JoshStrobl (commented):
Otherwise LGTM

PR #820, opened 2026-01-22
Add null checks for panel positioning to prevent crashes

fossfreedom:
## Description
As discussed on matrix, this PR adds a number of null checks in and around the panel positioning code.

For me this resolves @serebit TTY crash.

Definitely this code needs more testing around multimonitor and switching primary monitor as well as - I'm guessing pulling a monitor - the latter test case I'm not sure that we cater for currently - but hopefully this PR resolves potential crashes anyway.

### Submitter Checklist

### DIFF
--- a/src/panel/manager.vala
+++ b/src/panel/manager.vala
@@ -405,22 +405,22 @@ namespace Budgie {
 			if (primary == null) {
 				warning("Primary monitor not found (primary_monitor=%d). Cannot reposition panels.", primary_monitor);
 
-				// This in theory shouldn't be reached - but we probably should be safe
-				// than sorry here - lets be helpful and try to find any valid screen as fallback
-				if (screens.size() > 0) {
-					var first_screen = screens.lookup(0);
-					if (first_screen != null) {
-						warning("Using first available screen as fallback");
-						primary = first_screen;
-						primary_monitor = 0;
-					} else {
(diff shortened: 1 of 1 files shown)


src/panel/manager.vala
```diff
@@ -391,11 +402,37 @@ namespace Budgie {
...
+						primary = first_screen;
+						primary_monitor = 0;
+					} else {
+						warning("No valid screens available. Skipping panel repositioning.");
+						return;
+					}
+				} else {
+					warning("No screens in hashtable. Skipping panel repositioning.");
+					return;
+				}
```
JoshStrobl: Reduce nesting

```suggestion
				warning("Primary monitor not found (primary_monitor=%d). Cannot reposition panels.", primary_monitor);

				if (screens.size() == 0) {
					warning("No screens in hashtable. Skipping panel repositioning.");
					return;
				}

				// This in theory shouldn't be reached - but we probably should be safe
				// than sorry here - lets be helpful and try to find any valid screen as fallback
				var first_screen = screens.lookup(0);
				if (first_screen == null) {
					warning("No valid screens available. Skipping panel repositioning.");
					return;
				}

				warning("Using first available screen as fallback");
				primary = first_screen;
				primary_monitor = 0;
```

PR #822, opened 2026-01-24
Rework our brightness handling to not use settings-daemon

fossfreedom:
## Description
SettingsDaemon has moved most of its brightness handling into gnome-shell and as such the dbus handling we previously needed is not available.
In 10.9.x we used the GNOME 48 cut down version to handle matters - this doesn't work in wayland since it was X-Server specific.

So this PR replaces settingsdaemon with calls to logind where we can also set brightness - the complicating factor is that to get levels we have to consult the /sys based brightness files.

Note the inspiration statement in the brightness-manager.vala where the vast majority of all of this logic was found in various projects.  No code was copied and anyway this has been written in vala.

### Submitter Checklist

### DIFF
--- a/src/daemon/brightness-manager.vala
+++ b/src/daemon/brightness-manager.vala
@@ -7,12 +7,6 @@
 * it under the terms of the GNU General Public License as published by
 * the Free Software Foundation; either version 2 of the License, or
 * (at your option) any later version.
-*
-* Note: the following projects were consulted as inspiration for the
-* logic decisions
-* Obtaining XDG_SESSION_ID  - systemd project documentation
-* How to use loginD SetBrightness - wlroot library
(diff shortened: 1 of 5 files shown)


src/daemon/brightness-manager.vala
```diff
@@ -0,0 +1,343 @@
...
+		}
+		
+		/**
+		* Read current brightness from sysfs
+		*/
+		private void update_current_brightness() {
+			if (backlight_path == null) return;
+			
+			try {
+				var brightness_file = File.new_for_path(backlight_path + "/brightness");
```
JoshStrobl:
```suggestion
				var brightness_file = File.new_for_path(Path.build_filename(backlight_path, "/brightness"));
```

PR #825, opened 2026-01-25
fix: tr.po

YusufErdemK:


fossfreedom:
Many thanks for this. However, the correct place for translations is via transifex. We will then pull in updates as part of a release.

Do consider joining and helping. 

https://explore.transifex.com/buddiesofbudgie/

PR #827, opened 2026-01-28
Launch pkexec apps with wayland info - closes #777

### LINKED ISSUE #777: [Bug] menu applet: For pkexec based apps under wayland need to pass wayland details
### Description

Where we have used "pkexec" in the code for the menu applet we need to supplement something like the following to allow the app to run successfully under wayland

    pkexec env WAYLAND_DISPLAY="$XDG_RUNTIME_DIR/$WAYLAND_DISPLAY"  XDG_RUNTIME_DIR=/user/run/0  lightdm-settings

e.g.

    pkexec env WAYLAND_DISPLAY="$XDG_RUNTIME_DIR/$WAYLAND_DISPLAY"  XDG_RUNTIME_DIR=/user/run/0  lightdm-settings

### Budgie version

10.10-preview.1

### Operating System

UB 26.04

Launch any application from the menu where the Exec= in the .desktop file contains pkexec to launch the app with elevated permissions

### Actual result

App doesnt start - look in journalctl and note the log states it couldn't find display: 0

### Expected result

App starts with elevated permissions

### Additional information

_No response_

fossfreedom:
## Description
Rework how we handle pkexec based launches from the menu

### Submitter Checklist

### DIFF
--- a/src/lib/application.vala
+++ b/src/lib/application.vala
@@ -110,91 +110,66 @@ namespace Budgie {
 		public bool launch_with_context(AppLaunchContext context) {
 			try {
 				var info = new DesktopAppInfo(this.desktop_id);
+				var cmd  = info.get_commandline();
 
-				var cmd = info.get_commandline();
 				string[] parsed_args;
-				bool fallback = false;
+				GLib.Shell.parse_argv(cmd, out parsed_args);
 
-				try {
-					GLib.Shell.parse_argv(cmd, out parsed_args);
-				} catch (Error e) {
-					warning("Failed to parse command line for '%s': %s", this.desktop_id, e.message);
-					// Fallback: try the default launcher
-					fallback = true;
-				}
-
-				// If not pkexec, use the DesktopAppInfo to launch the app normally
-				if (parsed_args.length == 0 || parsed_args[0] != "pkexec" || fallback == true) {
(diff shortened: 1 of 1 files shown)


src/lib/application.vala
```diff
@@ -110,40 +110,97 @@ namespace Budgie {
...
+				while (i < parsed_args.length && parsed_args[i].has_prefix("-")) {
+					pkexec_opts.append(parsed_args[i]);
+					i++;
+				}
+
+				// The remainder after options is the original executable and its arguments
+				var original_prog_and_args = new GLib.List<string>();
+				for (int j = i; j < parsed_args.length; j++) {
+					original_prog_and_args.append(parsed_args[j]);
+				}
```
JoshStrobl: These aren't used until the iteration on line 174 over to append to final args. Instead of creating an array only to loop over it to append to another array, how about just doing that on L174 and directly appending to final_args / string array (more details below)? We aren't manipulating original_prog_and_args or parsed_args between those two points.

Also given we are appending final_args then to a newly constructed string array, would it maybe be better to instead create that string array earlier on, at least at L161 if not earlier, and push the WAYLAND_DISPLAY and XDG_RUNTIME_DIR directly to that, and then the "pkexec", then all the args?

PR #829, opened 2026-01-29
fix(panel): alignment of panel in dock modes, dynamic positioning

JoshStrobl:
## Description
This previously was not a problem as the window and main_layout allocations were the same and we were handling the positioning ourselves, but with layer shell that isn't necessary and the surface gets the full allocation, and the inner main_layout is the "real" smaller panel.

This also attempts to fix the issue with panel not hiding when a window is maximized. I tried to get it functioning in a manner where the menu (activated action) triggers a layer change, but it's still finnicky (YMMV).

### Submitter Checklist

### DIFF
--- a/src/daemon/screenlock.vala
+++ b/src/daemon/screenlock.vala
@@ -258,8 +258,11 @@ namespace Budgie {
 			int idle_brightness_percent = power.get_int("idle-brightness");
 			debug("Setting brightness to %d%%", idle_brightness_percent);
 
-			// Set to idle brightness
-			brightness_manager.set_brightness_percent((double)idle_brightness_percent / 100.0);
+			int current_brightness_percent = (int)(brightness_manager.get_brightness_level() * 100.0);
+			if (current_brightness_percent > idle_brightness_percent) {
+				// Set to idle brightness only if the current level is more than the dim level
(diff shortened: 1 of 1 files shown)


JoshStrobl:
Marked as draft. While it works for horizontal panels, need to fix it for verticals.

JoshStrobl:
Layer change being non-functional may be a compositor issue, we are making the correct calls (via GtkLayerShell) but no change:

```
[ 542631.651] {Default Queue}  -> zwlr_layer_surface_v1#55.set_exclusive_zone(36)
[ 542631.655] {Default Queue}  -> zwlr_layer_surface_v1#55.set_layer(3)
```

fossfreedom:
Some initial testing:

image

Top panel on login - move the panel right and the label on the left doesnt get updated until you close & open settings.

After than the create panel button doesn't do anything

Maximise a window such as firefox on the right and there is a couple of pixel separation - not theme dependent

image

So intellihide hides the dock in maximised - good. Autohide doesn't do anything on maximise though i.e. still behaves like a static panel

As you have noted - difficult to get the panel to reappear - workaround would be to press Super to open the menu. But closing the menu doesnt make the panel hide again.

PR #830, opened 2026-01-30
Extend the command line reset to include raven

fossfreedom:
## Description
Seems we don't actually reset the raven schema.

Particularly important for UB since ShowTime is now a raven widget as well as workspace-wallpaper-switcher widgets.

This PR adds a --reset-raven option that can be used separately or in-conjunction with --reset

### Submitter Checklist


JoshStrobl (approved):
LGTM, thanks!

PR #835, opened 2026-02-01
Tidy up bridge

fossfreedom:
## Description
Changes in this PR includes handling defining multiple rc.xml xml keys where there are more than one key in the keybind array key.
Also rc.xml keybinds now are fully linked against schema keys to allow dconf/BCC changes to reflect through the bridge. Lastly - our show menu Super_L which is linked to the mutter-overlay key is now overridden - in GNOME this is normally Super whereas in Budgie we are expecting Super_L so use a schema override here

~~marking as draft to-do more testing~~

### Submitter Checklist

### DIFF
--- a/src/bridges/labwc/labwc_bridge.py
+++ b/src/bridges/labwc/labwc_bridge.py
@@ -888,6 +888,21 @@ class Bridge:
 
             replacement = replacement.replace("XF86XF86", "XF86")
 
+            # Convert the final key (non-modifier) to lowercase
+            if "-" in replacement:
+                parts = replacement.rsplit("-", 1)
+                if len(parts) == 2:
+                    modifiers = parts[0]
+                    key = parts[1]
+                    # Only lowercase if it's not an XF86 key or special key
+                    if not key.startswith("XF86") and len(key) == 1:
+                        key = key.lower()
+                    replacement = modifiers + "-" + key
+            else:
+                # No modifiers, just a single key
+                if len(replacement) == 1:
+                    replacement = replacement.lower()
+
         return replacement
 
     # all keybinds from various gsettings schemas are managed


jls17:
Somehow this version of the labwc_bridge does break a lot of my media-keys.

That's a partial diff .. on the right-side the bridge did change many keys to "undefined".

```diff
$ diff rc.xml tidy-bridge-rc.xml
2c2
<       <desktops number="4" >
---
>       <desktops number="4">
34c34
<               <keybind bridge="plugins.media-keys/home" key="XF86Explorer">
---
>               <keybind bridge="plugins.media-keys/home" key="undefined">
37c37
<               <keybind bridge="plugins.media-keys/calculator" key="XF86Calculator">
---
>               <keybind bridge="plugins.media-keys/calculator" key="undefined">
40c40
<               <keybind bridge="plugins.media-keys/email" key="XF86Mail">
---
>               <keybind bridge="plugins.media-keys/email" key="undefined">
46c46
<               <keybind bridge="plugins.media-keys/www" key="XF86WWW">
---
...
...
199c253
<               <keybind bridge="plugins.media-keys/stop" key="XF86AudioStop">
---
>               <keybind bridge="plugins.media-keys/stop" key="undefined">
202c256
<               <keybind bridge="plugins.media-keys/volume-down" key="XF86AudioLowerVolume">
---
>               <keybind bridge="plugins.media-keys/volume-down" key="undefined">
205c259
<               <keybind bridge="plugins.media-keys/volume-mute" key="XF86AudioMute">
---
>               <keybind bridge="plugins.media-keys/volume-mute" key="undefined">
208c262
<               <keybind bridge="plugins.media-keys/volume-up" key="XF86AudioRaiseVolume">
---
>               <keybind bridge="plugins.media-keys/volume-up" key="undefined">
...
```

The previous version didn't touch those entries at all, because in line 756 the `keybindings_changed` method did return early.

In the newer version the `rc.xml` is overwritten during startup and those entries are all reset to "undefined". As a result I'm unable to use media-keys to control volume, start a calculator, and so on.

_What I can't tell_, how those media-key-entries got into the `rc.xml` file in the first place. I didn't configure all those keys manually, that's for sure. And the files under `/usr/share/budgie-desktop/labwc` don't have those key-assignments either (they are "undefined" there as well). (I must be missing something)

fossfreedom:
Move aside your current rc.xml, logout and login to allow the new rc.xml to be copied into place.

jls17:
That doesn't help. It re-creates the rc.xml from a template, but those keys aren't set anymore ...

```bash
$ grep plugins.media-keys/volume-up rc.xml
                <keybind bridge="plugins.media-keys/volume-up" key="undefined">
```

fossfreedom:
Use dconf-editor and look at your media keys schema keys. What is your volume up key set to?

jls17:
Now I did recreate the `rc.xml` while the previous version of the labwc_bridge was active.. result:

```bash
$ grep volume rc.xml
                <keybind bridge="plugins.media-keys/volume-down" key="XF86AudioLowerVolume">
                        <action name="Execute" command="wpctl set-volume @DEFAULT_AUDIO_SINK@ 5%-" />
                <keybind bridge="plugins.media-keys/volume-mute" key="XF86AudioMute">
                <keybind bridge="plugins.media-keys/volume-up" key="XF86AudioRaiseVolume">
                        <action name="Execute" command="wpctl set-volume @DEFAULT_AUDIO_SINK@ 5%+" />
```

The issue is the following code-block...

src/bridges/labwc/labwc_bridge.py
```diff
@@ -764,22 +774,71 @@ def keybindings_changed(self, settings, key):
...
+            for i, bridge in enumerate(existing_keybinds):
+                if i == 0:
+                    # Keep first element but set to undefined
+                    bridge.attrib["key"] = "undefined"
+                else:
+                    # Remove extra elements
+                    keyboard_element.remove(bridge)
+
+            self.write_config()
+            return
```
jls17: This block reset all those settings to "undefined" during labwc_bridge startup

fossfreedom:
Ah. I think i see the issue. Nice find. The intention here wasn't to tidy up all elements with at least one value. It was to tidy up when a key is emptied and all of the associated xml blocks were to be deleted. Should be an easy fix.

fossfreedom:
@jls17 Thanks for your feedback - before I re-mark this as ready, please can you retry the latest changes.
From my testing things like volume key was ok.  
I've also added a migration path - so the template keybinds in /usr/share/budgie-desktop/labwc will replace in your original local rc.xml.

jls17:
Just in case you want to include the following rc.xml fix:

```diff
diff --git a/src/bridges/labwc/rc.xml b/src/bridges/labwc/rc.xml
index 70681674..800db537 100644
--- a/src/bridges/labwc/rc.xml
+++ b/src/bridges/labwc/rc.xml
@@ -239,10 +239,10 @@
                        <action name="Maximize" direction="both" />
                </keybind>
                <keybind bridge="wm.keybindings/maximize-horizontally" key="undefined">
-                       <action name="Maximize" direction="right" />
+                       <action name="ToggleMaximize" direction="horizontal" />
                </keybind>
                <keybind bridge="wm.keybindings/maximize-vertically" key="undefined">
-                       <action name="Maximize" direction="left" />
+                       <action name="ToggleMaximize" direction="vertical" />
                </keybind>
                <keybind bridge="wm.keybindings/begin-move" key="A-F7">
                        <action name="" command="" />
@@ -257,7 +257,7 @@
                        <action name="Resize" />
                </keybind>
                <keybind bridge="wm.keybindings/unmaximize" key="W-Down">
-                       <action name="UnMaximize" direction="both" />
+                       <action name="UnSnap" />
                </keybind>
                <keybind bridge="wm.keybindings/toggle-fullscreen" key="undefined">
                        <action name="ToggleFullscreen" />
```

**What it does**

The `direction`-values in the `maximize-horizontally` and `maximize-vertically` bindings are invalid. Because of that, labwc does use `both` instead and performs a normal window-maximize. Additionally I like it, if I can press the binding twice and its simply toggling the window-state (so actions were switched to `ToogleMaximize`).

The second thing is change for W-Down from `UnMaximize` to `UnSnap`. I noticed, that once a window is snapped to a side, I can't use W-Down to bring it back to its normal state (pressing the key does nothing). Switching that action to "UnSnap" does work in both situations. It performs _UnMaximize_ and _UnSnap_ at the same time.

I know most things are just personal taste .. hopefully changing those makes sense. But if its kept unchanged, I can live with that as well.

fossfreedom:
@jls17 I've raise an issue to consider further your suggestion so that it doesn't get buried.

The lowercase issue was inherent in the original bridge code - so nice find.  Fixed it.

jls17:
Now test looks good on my end.

JoshStrobl (approved):
LGTM, thanks!

PR #847, opened 2026-02-07
Add background support for Crystal Dock

fossfreedom:
## Description
When autohiding/intellihiding, under wlroots compositors it does its trick via layershell - a 1 px border on the dock edge. For light backgrounds this leaves a noticeable bleed through when an app is maximised.  This PR detects when
crystal dock is running, and adds a 2px black edge border on the background being displayed disguising the issue. This is very dependent on the wallpaper sizing so the user needs to use the Style feature in BCC to ensure image ratio issues hide the actual edge being displayed.

UB will be offering crystal-dock as part of its solution - so this patch will be required to obviate the visual issue when using light-backgrounds

### Submitter Checklist

### DIFF
--- a/src/daemon/background.vala
+++ b/src/daemon/background.vala
@@ -13,17 +13,11 @@ namespace Budgie {
 	public const string BACKGROUND_SCHEMA = "org.gnome.desktop.background";
 	public const string ACCOUNTS_SCHEMA = "org.freedesktop.Accounts";
 	public const string GNOME_COLOR_HACK = "budgie-control-center/pixmaps/noise-texture-light.png";
-	public const string BACKGROUND_DBUS_NAME = "org.buddiesofbudgie.budgie.Background";
-	public const string BACKGROUND_DBUS_PATH = "/org/buddiesofbudgie/budgie/Background";
 
-	[DBus (name = "org.buddiesofbudgie.budgie.Background")]
 	public class Background : Object  {
 		private Settings? settings = null;
 
 		const int BACKGROUND_TIMEOUT = 850;
-
-		[DBus (visible = true)]
-		public signal void wallpaper_changing(string wallpaper_path, bool is_modified);
 
 		/* Ensure we're efficient with changed queries and dont update
 		* a bunch of times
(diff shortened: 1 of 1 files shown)


JoshStrobl:
Checking every 5 seconds for crystal dock seems a bit heavy handed. Couldn't we use Windowing to check if we have crystal dock (I imagine it has _some_ sort of app ID) and trigger changes (knowing if it is alive or not) based on our window_group_added / window_added / window_group_removed / window_removed events?

fossfreedom:
Quick check - crystal dock isn't picked up as  a top-level window - its not seen.  Guessing that its either filtered out by libxfce4windowing - or just not presented as a window via wayland stuff.

Had a look at if we could use dbus to connect - it doesnt present something we can use.

We can be smarter here  - firstly we shouldn't be doing anything if crystal-dock isn't installed - Environment.find_in_path.

I think it probably be wiser to gradually back away every 5 seconds - so increment every 5 seconds until - lets say - only check once every 30 seconds.  Think have to fiddle with removing the timeout and remaking it with the new value.

JoshStrobl:
It is likely being filtered out by the skip_tasklist check. If we made that configurable (when constructing the class), you could create a windowing class that doesn't make it it return early if the window is skipped, or we can shift the skip_tasklist logic to the consumers of windowing instead. Did you check that by any chance?

fossfreedom:
> It is likely being filtered out by the skip_tasklist check. If we made that configurable (when constructing the class), you could create a windowing class that doesn't make it it return early if the window is skipped, or we can shift the skip_tasklist logic to the consumers of windowing instead. Did you check that by any chance?

I checked the windows managed by the windowing class - deliberately not using the filters.  XFW only displays top-level windows.  crystal-dock (and other docks) are not top-level and are not registered by XFW.  Ho hum.

I've stripped out all the dbus stuff previously added to manage interaction with BDV.  That basically leaves the implementation as is - it gracefully extends its checks from 5 seconds to eventually once every 30 seconds which seems reasonable.

JoshStrobl:
Dang, alright. Will give this another look then. Thanks for checking!

PR #858, opened 2026-02-14
Workaround for qemu based issues with host acceleration (Closes #853)

### LINKED ISSUE #853: [Bug] upside down cursor in gnome-boxes / 3D graphics
### Description

Common issue apparently with gnome-boxes / qemu when you have 3D graphics enabled.  The mouse cursor appears upside down.

### Budgie version

10.10.1

### Operating System

UB 26.04

Launch an ISO in gnome-boxes - make sure you are full screen and have 3D acceleration enabled in gnome-boxes preferences

### Actual result

Note the cursor is upside down

### Expected result

Should be right way up

### Additional information

resolution is to edit /usr/bin/startbudgielabwc

add next to the virtualbox workaround

    WLR_NO_HARDWARE_CURSORS=1 
    export WLR_NO_HARDWARE_CURSORS

reboot

We need to wrap this up using something like this to test before exporting WLR_NO_HARDWARE_CURSORS

    glxinfo | grep "renderer string: virgl" to detect if we are using gnome-boxes/3D acceleration

fossfreedom:
## Description
Until qemu resolves matters (see #853) the best we can do is to tell wlroots not to use hardware cursors.  See the extra notes in #853 which is the sort of thing that needs to be in release notes

### Submitter Checklist

### DIFF
--- a/src/bridges/labwc/labwc_bridge.py
+++ b/src/bridges/labwc/labwc_bridge.py
@@ -27,6 +27,144 @@ from gi.repository import Pango
 
 mainloop = None
 
+CURRENT_RC_VERSION = 1
+
+class RcXmlMigration:
+    """Handles migration of rc.xml to newer versions"""
+
+    def __init__(self, bridge):
+        self.bridge = bridge
+        self.log = bridge.log
+
+    def get_rc_version(self, et):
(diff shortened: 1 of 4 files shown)


JoshStrobl (approved):
Thanks!

PR #859, opened 2026-02-14
Sync touchpad scrollmethod from gsettings

fossfreedom:
## Description
This updates the bridge to map the two gnome schema keys that control the touchpad scrollmethod to the scrollMethod config for labwc.

Since this would be missing from existing users rc.xml I've rejigged the bridge to create any peripheral key if they are missing.

Tested by deleting the entire devices section, running the bridge and seeing that all xml sections including the new scrollMethod is recreated.

### Submitter Checklist

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -49,7 +49,8 @@ jobs:
             slurp \
             swaybg \
             swayidle \
-            wlopm
+            wlopm \
+            egl-utils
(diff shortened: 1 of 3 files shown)


JoshStrobl (approved):
LGTM. Tested alongside BCC patch and rc.xml updates accordingly, scrolling behavior accurately reflected in application (tested with gnome-terminal)

Issue #852, opened 2026-02-10
[Request] Make labwc's environment file permament or don't override its set values


domesya:
### Budgie version

10.10.1

### Use-cases

Main use cases of having this file be permanent is to set more advanced keyboard options like `XKB_DEFAULT_OPTIONS=caps:backspace` to make the caps lock key behave as an additional backspace. Some other use cases are setting environment variables just for use with budgie. The file I'm referring to is located at `~/.config/budgie-desktop/labwc/environment` and currently gets overwritten on every login.

### Proposal

I'd propose to either make the file semi-permanent as in, make budgie-desktop-settings add to the file without fully overwriting it and only overwriting the appropriate variables, or add some form of a tool that allows adding custom variables and makes them permanent.

fossfreedom:
I have a pending PR that should address changing environment file content issues and accommodating user preferences.  It doesn't IMHO need a GUI to say "fixed or semi-fixed".

anyway - lookout for the PR when its done and please test.  Will tag this issue via the PR.

Note - needs this PR to be merged first since the changes are via the bridge. https://github.com/BuddiesOfBudgie/budgie-desktop/pull/835

jls17:
I miss that possibility too. In the PR the `XKB_DEFAULT_OPTIONS` are still hard-coded and overwritten by the bridge. For me, adjusting this setting is the only way to get Alt+Tab and Alt-Shift+Tab working at the same time. The `grp:alt_shift_toggle` interferes with the Alt+Shift+Tab keybinding. I dont switch keyboard layouts that often, so I'd like to make the layout-toggle use something else than Alt+Shift.

fossfreedom:
> In the PR the `XKB_DEFAULT_OPTIONS` are still hard-coded

I haven't pushed the PR yet.  #835 needs to be merged first before I submit the PR to resolve

PR #865, opened 2026-02-18
Rework our environment file handling

fossfreedom:
## Description
This hopefully add freedesktop locale1 support to the bridge, falls back to /etc/default/keyboard (if defined) and finally to "us".

So it should resolve calamares issues - that I can't develop or test with.

The environment file no longer gets rewritten - it merges user made changes, but writes bridged controlled items such as the locale.

Should hopefully mean we don't need to use gsd-keyboard as per #813 

Closes https://github.com/BuddiesOfBudgie/budgie-desktop/issues/852

### Submitter Checklist

### DIFF
--- a/src/bridges/labwc/labwc_bridge.py
+++ b/src/bridges/labwc/labwc_bridge.py
@@ -19,7 +19,6 @@ import psutil
 import sys
 import gettext
 from enum import StrEnum
-from gi.repository import Gio
 import dbus
 import dbus.mainloop.glib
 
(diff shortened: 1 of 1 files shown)


jls17:
I tested this version and noticed the following ...

- the previous version did write `XKB_DEFAULT_OPTIONS=grp:alt_shift_toggle` into the environment file
- the new version does read this from the existing environment file
- ... and it merges that setting with the values it retrieves from systemd-localed
- in my case the result was the invalid configuration: `XKB_DEFAULT_OPTIONS=grp:alt_shift_toggle,grp:shift_caps_toggle`

~~So the layout toggle didn't work at all.~~ After I manually removed the `grp:alt_shift_toggle` from the line, it was kept correctly. The same scenario happens, if the user is going to change a setting in the future. Or if setting get removed from systemd-localed ... its still kept in the environment file.

There should be only one option per option-category (in the example `grp:`). Maybe its not even necessary to read those values from the existing environment file, if they are configured in other places. But I couldn't really get my head around it .. not sure about it.

---
Another thing: I've set the systemd-localed to only "de" keyboard layout. In BCC I've added "en" as  secondary keyboard layout. And sometimes I reverse the order in BCC. The setting in BCC is now ignored completely.

Update: Now I did understand .. the toggle didn't work because there was only one layout.

fossfreedom:
Well we need to decide on the algorithm here and how todo the merge. We have 3 potential sources of info ... localed, /etc/default/keyboard and BCC. I am open to suggestions. I am not a multi language and layout person so I am open to suggestions.

For the toggle we need to decide on the master. So from what you said I think localed should be the master for any "toggle" except if that toggle is not a "led toggle". If not defined then use alt shift toggle as a fallback.

Maybe we should be assuming BCC is the master and if defined use it. Otherwise switch to using localed and then /etc/default/keyboard. That means you have to define both de and en in BCC

fossfreedom:
@jls17 the bridge has now been reworked - xkb-options is now fully under the bridge control - we no longer try to modify user defined changed. If the user wants to enter their own xkb-options then they should use the dconf key xkb-options under input-sources.

for both the layout and xkb-options they are both calculated from 

1. gsettings if defined 
2. locale1
3. /etc/default/keyboard
4. use a default value

Please give this a try and let me know if there are any issues how this is now functioning.  TIA

jls17:
This works really well for me.

- I can use/set a __single__ system-wide layout (using `localectl`) which gets used in case I don't start Budgie
- I can set xkb-options (using `localectl`)
- I can use BCC to add additional layouts and order them as needed

... and its all correctly reflected in labwc's environment file.

Here are my locale1-settings:

```
$ localectl
System Locale: LANG=de_DE.UTF-8
               LANGUAGE=de_DE
    VC Keymap: (unset)
   X11 Layout: de(nodeadkeys)
    X11 Model: pc105
  X11 Options: grp:shift_caps_toggle,grp_led:caps,caps:none
```

In BCC I have added the `en` keyboard as secondary layout. And that's the generated `environment` file:

```
$ cat environment
# Budgie Desktop - labwc environment configuration
# Variables fully managed by budgie: XKB_DEFAULT_LAYOUT, XKB_DEFAULT_OPTIONS, XCURSOR_*, LC_*, LANG
# Use dconf key xkb-options to add user defined values
# Other user customizations are preserved

XKB_DEFAULT_LAYOUT=de(nodeadkeys),us
XKB_DEFAULT_OPTIONS=caps:none,grp:shift_caps_toggle,grp_led:caps

XCURSOR_SIZE=24
XCURSOR_THEME=capitaine-cursors-light

LANG=de_DE.UTF-8

# User customizations
LANGUAGE=de_DE
```

Thank you very much

EbonJaeger:
And thank you, @jls17, for the additional testing!

PR #868, opened 2026-02-19
Fix: Close windows via popover

### LINKED ISSUE #867: [Bug] Panel - Closing windows via popover is buggy
### Description

If there are multiple instances of an app open, the button-popover contains a list of the opened instances. Closing a instance via the "X"-button of a line is buggy. It is correctly working if you close them in order (top to bottom); but if you close the last item first, it removes the wrong WindowItem-widget, leaving a unusable WindowItem-widget in the list view.

https://github.com/user-attachments/assets/fe63516c-41ff-45a9-a10f-d067c6e67fcc

### Budgie version

### Operating System

ArchLinux

1. open multiple instances of the same app (same IconButton)
2. right-click on the IconButton to open the popover
3. close the instances from bottom-to-top order 
4. for at least one line its not possible to close the related window

### Actual result

- Popover removes wrong item from list, thus it keeps invalid entry which can't be closed

### Expected result

- Popover should remote the corrent item from list
- it should be possible to close all the open windows via the popover in any order

### Additional information

_No response_

jls17:
## Description

The PR fixes #867 

---
1) Removing the correct ListView-item

Till now the items (list of open windows) in the popover were identified by the class_id (e.g. "nemo" or "code-oss"). But the class_id isn't unique per window. Because of that its not possible to find the correct unique window-entry in the list-view. All entries have the same identical class_id.

Now in the PR each listview-entry gets casted to `WindowItem`. After that the stored window-property is available and makes comparing the closing-window with the stored window in `WindowItem` possible. If it finds the correct item within the ListView, it gets removed/destroyed.

---
2) WindowControls

The `WindowControl`s are only shown if the user clicks on the arrow of a ListView-item. In that case a extra named-child gets added to the stack that will show controls like Maximize and Minimize. Till now the class_id was used as the stack-child's name. But the class_id is the same for all windows in the group, and thus it was not possible to use the class_id to identify the currently opened `WindowControl`'s window (e.g. within `remove_window`).

Because there was always only one WindowControls-pane open, the child-name within the stack was now changed to "controls". If a window of that ButtonGroup gets closed and this window's "controls"-pane is active, the stack is switched back to "main" and the active controls-pane gets destroyed.

---
Please let me know if I should improve something.

### Submitter Checklist

### DIFF
--- a/src/panel/applets/icon-tasklist/widgets/ButtonPopover.vala
+++ b/src/panel/applets/icon-tasklist/widgets/ButtonPopover.vala
@@ -196,39 +196,29 @@ public class IconTasklistButtonPopover : Gtk.Popover {
 				controls_layout.destroy();
 			});
 
-			var class_ids = window.get_class_ids();
-			if (class_ids == null || class_ids.length == 0) {
-				warning("Window '%s' has no class IDs", window.get_name());
-				return;
+			var control = stack.get_child_by_name("controls");
+			if (control != null) {
+				stack.set_visible_child_name("main");
+				stack.remove(control);
+				control.destroy();
 			}
 
-			string id = class_ids[0];
-			stack.add_named(controls_layout, id);
-			stack.set_visible_child_name(id);
+			stack.add_named(controls_layout, "controls");
+			stack.set_visible_child_name("controls");
 		});
 
 		windows.add(window_item);
 	}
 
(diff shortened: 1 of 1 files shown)


JoshStrobl (approved):
LGTM. Closing last item now works consistently, and no issues with the behavior of maximize / minimize for the windows.

Thank you very much! :)

PR #869, opened 2026-02-21
Ensure budgie-desktop is part of XDG_DATA_DIRS in the same way as the prior Budgie:GNOME session id

fossfreedom:
## Description
in budgie 10.9.x XDG_DATA_DIRS contained /usr/share/budgie-desktop.

With our wayland config this has been lost - so lets readd it.  This ensures things like the menu picks up /usr/share/budgie-desktop/applications before /usr/share/applications - i.e. distros can override desktop entries such as OnlyShowIn.  For example org.gnome.PowerStats.desktop is defined under Debian as GNOME;Unity and thus is never visible in Budgie.  By placing the equivalent .desktop in /usr/share/budgie-desktop/applications with a OnlyShowIn=Budgie; then it will appear.  It means therefore we don't have to start chasing various package managers to get Budgie included.

### Submitter Checklist

### DIFF
--- a/src/session/startbudgielabwc.in
+++ b/src/session/startbudgielabwc.in
@@ -33,6 +33,7 @@ if [ -z "${XDG_RUNTIME_DIR}" ]; then
         fi
     fi
 fi
+
 if [ ! -e "${XDG_RUNTIME_DIR}" ]; then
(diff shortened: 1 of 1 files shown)


JoshStrobl (approved):
LGTM, thanks!

PR #872, opened 2026-02-22
fix(session): ensure xdg-desktop-portal is started alongside our session

JoshStrobl:
## Description
This was done as part of the launch script as xdg-desktop-portal doesn't provide a correct signal to inform budgie-session that it has started, therefore it hangs session startup.

Having it in the launch script enables us to launch it with effectively no hangup before then launching the session.

Prior to this, xdg-desktop-portal would not be started as its dependencies are not fulfilled during our start-up.

### Submitter Checklist


fossfreedom (approved):
Entirely reasonable!

PR #878, opened 2026-02-24
Separate touchpad left-handed setting from mouse if requested - closes #877

### LINKED ISSUE #877: [Bug] Touchpad left-handed setting is confused
### Description

Looking at the bridge code, the left-handed setting for a touchpad is assumed to be a "yes" / "no" type of value.  That is wrong - the gsettings value is "mouse", "left", "right".  Thus the bridge code is mapping rc.xml peripherals section for the touchpad left-handed setting incorrectly.

### Budgie version

10.10.1

### Operating System

UB 26.04

In BCC touchpad fiddle with the left handed switches and examine the value in the rc.xml file.

### Actual result

rc.xml peripherals touchpad section does not reflect what the touchpad left-handed setting

### Expected result

rc.xml should respect if the touchpad is following the mouse settings or individually set.

### Additional information

_No response_

fossfreedom:
## Description
  This is a logic change since we incorrectly assume touchpad and mouse left/right settings are the same.
Labwc allows separation - so this PR provides this.

Note - there isn't actually a setting in BCC for this - so its a behind the scene ability by directly poking at gsettings

### Submitter Checklist

### DIFF
--- a/src/bridges/labwc/labwc_bridge.py
+++ b/src/bridges/labwc/labwc_bridge.py
@@ -808,7 +808,6 @@ class Bridge:
             category = "non-touch"
 
         yesno = { "natural-scroll" : "naturalScroll",
-                  "left-handed" : "leftHanded",
                   "accel-profile" : "accelProfile",
                   "tap-to-click" : "tap",
                   "tap-and-drag" : "tapAndDrag",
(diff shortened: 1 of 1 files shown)


Zoddo:
FYI, applied commit 46ca7d285eef15e0ef69693d91652ce537c4e8c2 (for the function `ensure_peripheral_element`) + this PR on top of 10.10.1, and I confirm it fixes the inverted buttons for me :)

PR #894, opened 2026-03-14
Bdv vs crystal

fossfreedom:
## Description
Reported on our discourse forum - toggling the desktop icons option in BDS freezes crystal-dock.

Ordinarily for workaround type issues I would normally suggest to log out and login again. 

But freezes are ugly. This PR briefly stops crystal-dock and restarts it.  This has been confirmed as working.

The change is part of the crystal-doc-helper.vala and as such has no impact on non crystal-dock users.

### Submitter Checklist


EbonJaeger (approved):
LGTM, thanks!

PR #901, opened 2026-04-12
Allow our accounts-service interaction cope with the freedesktop dbus name as well as the lightdm variant

fossfreedom:
## Description
Noticed that the very latest accountsservice uses a standard freedesktop dbus name whereas budgie-desktop is expecting a lightdm variant of the dbus name.

This PR ensures we can cope with both variants - new accounts service and legacy lightdm based dbus naming

### Submitter Checklist

### DIFF
--- a/src/daemon/background.vala
+++ b/src/daemon/background.vala
@@ -114,7 +114,6 @@ namespace Budgie {
 		void set_accountsservice_user_bg(string background) {
 			DBusConnection bus;
 			Variant variant;
-
 			try {
 				bus = Bus.get_sync(BusType.SYSTEM);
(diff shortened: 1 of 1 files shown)


src/daemon/background.vala
```diff
@@ -114,14 +114,14 @@ namespace Budgie {
 		void set_accountsservice_user_bg(string background) {
 			DBusConnection bus;
 			Variant variant;
-
```
EbonJaeger, line 117: Just wondering why this line was removed? Seems like unneeded patch diff.

EbonJaeger (approved):
LGTM, thanks!

PR #906, opened 2026-04-20
build: Fix missing dep for pkgconfig and GIR

EbonJaeger:
## Description

Fixes GtkLayerShell being missing from the dependency list when generating the GIR and pkgconfig files for panel plugin.

### Submitter Checklist


JoshStrobl (approved):
LGTM, thanks!

PR #911, opened 2026-04-21
tasklist: Fix minimize cases

### LINKED ISSUE #907: Budgie 10.10.2: can't minimize some apps by clicking on Tasklist item
Budgie 10.10.2 on Solus.

Can't minimize some apps by clicking on Tasklist item (not icon-tasklist). It does nothing (but item color is changing).
Affected apps: firefox, vivaldi

EbonJaeger:
## Description

This makes the logic flow more clear, and also fixes the check for if a window is activated. Previously, we checked if the entire state was equal to `WindowState.ACTIVE`, which is incorrect because window state is a set of bit flags.

Fixes https://github.com/BuddiesOfBudgie/budgie-desktop/issues/907

### Submitter Checklist


JoshStrobl (approved):
LGTM!

PR #921, opened 2026-05-04
Rework media keys and other rc.xml defaults handling

### LINKED ISSUE #775: [Bug] dconf setting breaks custom keyboard shortcut on reboot
### Description

I have always used a custom keyboard shortcut for `Ctrl+Alt+T` to launch my terminal. For a while now (from the release notes I would assume since 10.9.3) this shortcut stopped working after rebooting. It is still present in the list of custom keyboard shortcuts but not working. Resetting the shortcut enables it again until the next reboot.

It turns out this is related to the `dconf` setting `org.buddiesofbudgie.settings-daemon.plugins.media-keys.terminal`. By default this is set to `['<Primary><Alt>t']`.

From the description *Binding to launch the terminal.* I assume this is intended to perform the same function as my  custom shortcut but doesn't yet do so. At least on my system without the custom shortcut `Ctrl+Alt+T` does nothing.

I used `dconf-editor` to change the value to an empty string and now my custom shortcut is kept intact after rebooting.

### Budgie version

10.9.4

### Operating System

Solus 4.8

1. Add custom keyboard shortcut for `Ctrl+Alt+T`. 
2. Reboot

### Actual result

Shortcut stops working after reboot.

### Expected result

Shortcut persists after reboots.

### Additional information

_No response_

### LINKED ISSUE #839: [Request] Budgie 10.10.1: option to disable OSD when switching workspaces
### Budgie version

10.10.1

### Use-cases

In Budgie 10.10 when you switch workspaces you will see OSD info in the center of the screen.

- it could be distracting for some users
- there is no reason for it when you have workspaces applet on the panel
- there wasn't any OSD in previous versions

I suggest to disable OSD by default and add option to enable it in the Budgie Settings (Desktop section) below `Number of virtual desktops` item.

### Proposal

I suggest to disable OSD by default and add option to enable it in the Budgie Settings (Desktop section) below `Number of virtual desktops` item.

### References

_No response_

### LINKED ISSUE #916: [Bug] Keyboard Media keys don't work
### Description

Keyboard Media keys like Play/Pause and Next/Previous don't work

### Budgie version

10.10.2

### Operating System

Solus

Press Play/Pause on the keyboard when music player is playing (that can be controlled with global media keys)

### Actual result

nothing happens

### Expected result

Working keys

### Additional information

Volume keys do work

fossfreedom:
## Description
This updates labwc rc.xml with the following updates as we have learned over recent times

1. audio media keys so that multimedia keyboards have something to hang their keys on - uses playerctl
2. Switch to a distro agnostic method to launch the media keys - terminal keybind (CTRL + ALT + T extant)
3. Switch to a distro agnostic method to launch media key calculator
4. Use to a distro agnostic method to launch media key web-browser - note we now accommodate XF86HomePage as a media key
5. Use to a distro agnostic method to launch media key for file-manager
6. Use to a distro agnostic method to launch media key - media player e.g. vlc/lollypop etc
7. Add the desktops - popuptime entry to give people a hint that there is a setting here to configure - e.g. setting to zero hides the workspace OSD
8. Add numlock to give people a hint that this can be configured by end users i.e. set numlock on login
9. Add bridge components to allow people to define keybinds for workspaces 5 to 8

Closes: https://github.com/BuddiesOfBudgie/budgie-desktop/issues/916
Closes: https://github.com/BuddiesOfBudgie/budgie-desktop/issues/839
Closes: https://github.com/BuddiesOfBudgie/budgie-desktop/issues/775
Closes: https://github.com/BuddiesOfBudgie/budgie-desktop/issues/896

### Submitter Checklist


JoshStrobl:
I'd like to push back on adding playerctl as a dep. This hasn't seen any changes in over 5 years, you have to run playerctld for it to keep track of players (so another integration point), and there are swathes of bugs across both that have been left unaddressed upstream. We have a pre-existing, functional MprisClient and related code for MPRIS interface generation in Raven as-is, so it would be good to see that be re-used instead.

For example:

- Mpris iface definitions and setup get shifted out of media-controls into a lib/mpris.vala class. Add tracking of most recent client iface (I'd say based off play / pause / stop transitions). Client would expose that as, signals for new client / destroyed client that provide the MprisClient that is being created or destroyed.
- Leverage this class in MediaControlsRavenWidget (show all ifaces)
- Leverage this class in budgie-daemon for exposing DBus interfaces to play/pause/prev/next on current iface (if one exists)
- Use dbus-send to fire off dbus requests against our interface

fossfreedom:
Fair enough - certainly hell of a lot more invasive change than was envisaged.  Looking over the mpris widget think we can do a lot of code move out and substitute calls for the static library.  But much more file touches to worry over / test over.

JoshStrobl:
Yea I agree that it is a more invasive change and I certainly appreciate you taking the time to make this PR and test it. I'm just trying to be cognizant of adding in more runtime deps, especially those that aren't well maintained. It's great for quickly enabling features / functionality (similar to what we did with swaybg and co.) but when we already have the support in our code base for a given functionality, my personal _preference_ is we take the time to adapt it for the broader usecase.

JoshStrobl (approved):
LGTM!

PR #922, opened 2026-05-07
Add a brief delay to allow workspace spinner to complete - closes #903

### LINKED ISSUE #903: [Bug] Amount of Workspaces setting plus/minus buttons do not stop till either 1 or 8
### Description

In Budgie Desktop Settings if you change the amount of workspaces with the plus button it will increment by 1 till the max of 8, likewise with the minus button it will lower it down to 1.

I need to enter a number manual and then press Enter instead.

### Budgie version

10.10.2

### Operating System

Gentoo

1. Open Budgie Desktop Settings and go to the amount of workspaces setting
2. press plus button once

### Actual result

It increases from the number (e.g. 4) to 5, then 6, then 7 and stops at 8

### Expected result

Increases from the current to current +1.

### Additional information

Can't seem to stop it till it is done

fossfreedom:
## Description
Tidy up so odd code dealing with workspace changes in BDS.
Added a small delay to ensure that the button click on the Gtk Adjustment widget completes before calling labwc -r

Tested by repeatedly clicking the +- buttons faster than the 0.5 second delay and the workspaces correctly sorted themselves out.

### Submitter Checklist

### DIFF
--- a/src/bridges/labwc/labwc_bridge.py
+++ b/src/bridges/labwc/labwc_bridge.py
@@ -1348,6 +1348,8 @@ class Bridge:
             updated = True
 
         if updated:
+            import time
+            time.sleep(0.5)
             self.write_config()
 
     # all gnome desktop interface gsettings changes are managed
--- a/src/bridges/labwc/labwc_bridge.py
+++ b/src/bridges/labwc/labwc_bridge.py
@@ -21,6 +21,7 @@ import gettext
 from enum import StrEnum
 import dbus
 import dbus.mainloop.glib
+import time
 
 import gi
 from gi.repository import Gio, GLib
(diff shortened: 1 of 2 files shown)


src/bridges/labwc/labwc_bridge.py
```diff
@@ -1342,6 +1342,8 @@ def desktop_wm_preferences_changed(self, settings, key):
             updated = True
 
         if updated:
+            import time
+            time.sleep(0.5)
```
JoshStrobl: shouldn't time be imported at the start of the file? Would this not re-import every time updated is true?

JoshStrobl (approved):
LGTM, thanks!

Issue #670, opened 2025-01-30
[Request] Open last entry in taskbar window group when clicked


denravonska:
### Budgie version

10.9.2

### Use-cases

When the same application, such as VScode, is opened multiple times they are grouped into one entry on the taskbar. Right clicking on that entry allows you to select which instance to bring to bring to the front, but clicking on the group itself will always open one specific instance.

### Proposal

Clicking on the group should open the instance that was last in focus.

denravonska:
It's weird because it looks like it's supposed to open [the last active window](https://github.com/BuddiesOfBudgie/budgie-desktop/blob/main/src/panel/applets/icon-tasklist/IconTasklistApplet.vala#L538), but it doesn't. It always opens the first window.

denravonska:
This seems to fix it. I don't know the code base enough to tell if it's the correct approach or not.

[last-active-window-fix.txt](https://github.com/user-attachments/files/28630028/last-active-window-fix.txt)

fossfreedom:
@denravonska was this tested on 10.9.2 or tested on 10.10?

denravonska:
> [@denravonska](https://github.com/denravonska) was this tested on 10.9.2 or tested on 10.10?

I tested this briefly against latest main (6cb2e0a319987df6b27c4f8046405d91dd6dd73e).
I'll apply it to my main workstation to let it run a bit more.

denravonska:
Ran it for a day on my dev machine under Wayland. So far it's been working fine and I cannot tell you how blissful it is to not making changes in the wrong vscode and not having to right click to select the correct instance :)

JoshStrobl:
Great to hear it is working for you :)

JoshStrobl:
@denravonska if you could submit a proper PR though that'd be appreciated rather than a diff that'd be great! That way it also makes it clear that you committed the code / contributed it.

denravonska:
> [@denravonska](https://github.com/denravonska) if you could submit a proper PR though that'd be appreciated rather than a diff that'd be great! That way it also makes it clear that you committed the code / contributed it.

Will do!

PR #899, opened 2026-03-29
docs: add contributing guidelines, AI policy, DCO, and agent docs

JoshStrobl:
## Description

Add human-centric development documentation for budgie-desktop:

- `CONTRIBUTING.md` with build quick reference (distro-specific), conventional commit types/scopes, DCO requirements, AI policy summary, and PR expectations
- `AGENTS.md` with technical context for AI coding assistants (project identity, repo structure, Wayland context, external resources, attribution rules, code conventions)
- `CLAUDE.md` as thin pointer to AGENTS.md and CONTRIBUTING.md
- `DCO.txt` with Developer Certificate of Origin v1.1
- Updated PR template with test plan section, DCO sign-off checklist, and AI policy checklist
- Added `/commit` and `/pr` Claude Code skills with budgie-specific configuration

Full AI policy and contributing docs live on the docs site — this PR references them rather than duplicating:
- https://docs.buddiesofbudgie.org/organization/ai-policy
- https://docs.buddiesofbudgie.org/developer/contributing

## Test plan

- [x] All cross-reference links resolve correctly

### Submitter Checklist

- [x] All commits are signed off (`git commit -s`) per the [DCO](DCO.txt)
- [x] Squashed commits with `git rebase -i` (if needed)
- [x] Built budgie-desktop and verified that the patch worked (if needed)
- [x] If AI tools or LLMs were used as part of this contribution, I have read the [AI Policy](https://docs.buddiesofbudgie.org/organization/ai-policy) and commits include `Assisted-by` trailers where applicable

PR #928, opened 2026-06-06
Raise the last used window in a window group on tray clicks.

denravonska:
This fixes an issue where clicking on a multi-window application in the tray would always open an arbitrary (first opened?) window regardless of what was opened before. This changes the behavior to always open the window that was the last focused window in the group.

It seems like this was always the intent but it wasn't working reliably on Wayland. May need more testing on X11.
This is a simplified version of the patch posted in #670.

Full disclosure: I don't know this code base so I used Claude to solve my problem. It is working for me and I tried to keep the diff down, but you guys should be able to tell if it's out of line or not.

### DIFF
--- a/src/windowing/window_group.vala
+++ b/src/windowing/window_group.vala
@@ -24,8 +24,10 @@ namespace Budgie.Windowing {
 		public DesktopAppInfo? app_info { get; construct; default = null; }
 
 		private List<unowned Xfw.Window> windows;
+
+		// The most recently focused window in the group. Deliberately not cleared
+		// when focus leaves the group, so it always holds the last window used.
 		private unowned Xfw.Window? active_window = null;
-		private unowned Xfw.Window? last_active_window = null;
 
 		/**
 		 * Emitted when the active window in the group changes.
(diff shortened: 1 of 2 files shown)


JoshStrobl:
> Full disclosure: I don't know this code base so I used Claude to solve my problem. It is working for me and I tried to keep the diff down, but you guys should be able to tell if it's out of line or not.

No worries, but please follow our AI Policy then and make sure to have the Assisted-by and sign off the commit!

- https://docs.buddiesofbudgie.org/organization/ai-policy
- https://docs.buddiesofbudgie.org/developer/contributing

(I won't blame your Claude for not knowing about this though, since #899 isn't merged in yet).

> May need more testing on X11.

We don't support X11 in Budgie 10.10 so this is a non-issue.

denravonska:
> No worries, but please follow our AI Policy then and make sure to have the Assisted-by and sign off the commit!
> 
> * https://docs.buddiesofbudgie.org/organization/ai-policy
> * https://docs.buddiesofbudgie.org/developer/contributing
> 
> (I won't blame your Claude for not knowing about this though, since #899 isn't merged in yet).

Fixed!

JoshStrobl (approved):
LGTM, thanks!

PR #934, opened 2026-06-12
Use static module variable for wayland client to stop freezes

fossfreedom:
## Description

This PR resolves a continuing issue with budgie-polkit-dialog where is can and does occasionally spin to 100% and polkit dialogs and popups like logout, shutdown and reboot no longer appears.

The issue is due to libxfce4windowing invoking  wayland client requests that seemingly lock each other out I.e. dont allow each to run to completion.

This PR ensures only one reference to wayland client (our wrapper around lfw) and thus requests to this dialog reference are guaranteed to be a singular rather than depending on internal lfw references cleaning themselves up when the dialog was previously releasing its reference 

## Test plan

Use gnome-disks where i can repeatedly reproduce this when writing isos to usb drives. I.e. polkit should be invoked.

### Submitter Checklist

### DIFF
--- a/src/dialogs/polkit/polkitdialog.vala
+++ b/src/dialogs/polkit/polkitdialog.vala
@@ -10,6 +10,7 @@
  */
 
 namespace Budgie {
+	static WaylandClient? wayland_client = null;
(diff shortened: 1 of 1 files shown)


JoshStrobl (approved):
LGTM, thanks!

PR #936, opened 2026-06-16
build: Add option to set xdp libexecdir

EbonJaeger:
## Description

On Solus, we set a package's libexecdir to `lib64/${package}` or `lib32/${package}`, which has the side effect of some of our build variables being incorrect. This option enables us to set the libexecdir for xdg-desktop-portal manually, thus avoiding the issue.

## Test plan

Rebuild the Solus 10.10 testing package with the new option, and verify that the xdp path in `/usr/bin/budgie-desktop` is correct.

### Submitter Checklist


JoshStrobl (approved):
LGTM, thanks!

PR #937, opened 2026-06-17
feat(keyboard-layout): support layout switching on wayland

### LINKED ISSUE #650: Implement Wayland support for Keyboard Layout applet
This applet currently causes segfaults under Wayland, likely due to how we use IBus, or the use of `Gnome.Languages` + the XCB related code.

EbonJaeger:
## Description

The keyboard layout applet did not work on Wayland. Switching a layout now goes through a new `org.buddiesofbudgie.KeyboardLayout` interface on budgie-daemon, which relays the request to whichever compositor bridge is running — the labwc bridge applies it by writing `XKB_DEFAULT_LAYOUT` and reloading the compositor. Keeping the daemon in the middle keeps this compositor agnostic: any bridge can implement the same contract, such as Ubuntu Budgie's Wayfire bridge. locale1's `SetX11Keyboard` is deliberately avoided, since some distributions block it.

Nothing reported back which layout was actually in use, so that interface also carries a `CurrentLayout` property. Bridges publish it — the labwc one reads it back from the environment file it writes — and the applet shows it as a label on the panel.

The applet itself was rewritten in C along the way. It keeps layout selection, adds a header button that opens the keyboard page in Budgie Control Center, and no longer goes through IBus — input source details come straight from libgnome-desktop.

When no input sources are configured at all, as on a live image or a fresh install, the applet derives one from the system locale rather than showing nothing.

Fixes #650

## Test plan

- Switch layouts from the applet and confirm the panel label follows.
- Change the layout elsewhere (`gsettings`, or D-Spy on locale1) and confirm the applet updates.
- Add, remove, and fully reset `org.gnome.desktop.input-sources`, watching `journalctl --user` for criticals.
- With no input sources configured, confirm the applet still shows a layout.
- Confirm the header button opens the keyboard settings page.

### Submitter Checklist


EbonJaeger:
Left to do:

- [x] I still need to document most of it.
- [x] Final cleanup pass.
- [x] We need to decide and implement how we want to handle this for labwc in the bridge. I proposed changing the bridge to only set the current layout instead of appending newly found ones, to hopefully force labwc to immediately switch to it.
      Another option might be to save a separate environment file for each layout, and use the bridge to swap them out when a layout is selected.

I'm not really familiar with the workings of the bridge, so input from @fossfreedom would be important.

EbonJaeger:
Final cleanup and documentation done.

fossfreedom:
[output.patch](https://github.com/user-attachments/files/29166204/output.patch)

The above fixes for me the immediate crash when no sources are defined.

Under labwc, the applet correctly states that my keyboard layout is English (UK).  Under wayfire it states US.  When I get some time I'll throw in some debugging to find out why.

fossfreedom:
@EbonJaeger before fiddling with the bridge, do note the bridge already responds to the keyboard layouts listed in budgie-control-center - so if you change the order here, the bridge updates XKB_DEFAULT_LAYOUT.

So maybe the applet should just update that gsettings key to reflect the updated order - i.e. the selected applet keyboard first followed by the other layouts.

Thoughts on this approach?

fossfreedom:
@EbonJaeger made some changes to this via a branch https://github.com/BuddiesOfBudgie/budgie-desktop/commit/6da18a0f7574a6edc770a51a20e64a0635cd5bc5

This introduces a dbus interface as per standup that when called by the applet, signals the bridge to update the keyboard layout - so in the case of labwc, update the environment file.

Feel free to hack-away.

JoshStrobl:
DCO is busted because I had used the GitHub feature for batch adding suggestions which makes a commit. It signed it with `Signed-off-by: Joshua Strobl <JoshStrobl@users.noreply.github.com>`

I clicked the button in DCO to "manually set to pass".

JoshStrobl (approved):
LGTM 🚀

PR #938, opened 2026-06-20
Add labwc raiseOnFocusDelay for the focus-mode - closes #750

### LINKED ISSUE #750: [Bug] Using focus follow mouse (both `sloppy` and `mouse` mode) causes some dropdown to not be interactable
### Description

If you have set with `org.gnome.desktop.wm.preferences focus-mode` to `mouse` or `sloppy`, some dropdowns are not interactable, and clicks are sent to what's behind it instead. 

Notably, I have this happened on Steam where if I right click a game and try click "properties" and similar, the pop down just closes and the click registers on the UI elements behind it. Currently I have to change back to non-focus follow mouse mode with `gsettings set org.gnome.desktop.wm.preferences focus-mode 'click'` to click the element and then re-enable focus follow mouse.

I have tried change both `org.gnome.desktop.wm.preferences auto-raise-delay` and `org.gnome.desktop.wm.preferences auto-raise` around but that does not seem to be contributing to this issue and it happens regardless of what I have for those settings

https://github.com/user-attachments/assets/aed87f85-7d56-4ec1-8884-b17b461a028a

### Budgie version

10.9.2

### Operating System

Ultramarine Linux 42 (Flagship Edition)

1. Run either ``gsettings set org.gnome.desktop.wm.preferences focus-mode 'mouse'`` or ``gsettings set org.gnome.desktop.wm.preferences focus-mode 'sloppy'`` 
2.  Open Steam(Currently only found this issue happening with Steam dropdowns which is why I am using it here) 
3. Right click on a game that is on the left panel of the UI, which should popup options (add to favorites, properties, etc)
4. Left clicking on any of the option.

### Actual result

The dropdown disappeared, the left clicked option was not registered on the dropdown but on what's behind it instead.

### Expected result

The dropdown disappeared and open its respective option.

### Additional information

_No response_

fossfreedom:
## Description

bridge support for labwc raiseOnFocusDelay - use with Mouse FocusMode to delay switching

## Test plan

labwc 0.20
start the bridge - confirm new raiseonfocusdelay key was created in the focus xml section

switch focus-mode to mouse and open two windows - move your mouse out of one window and within the bridge key value (500ms by default) switch back. Confirm that windows hasn't automatically been raised.

Now repeat but move the mouse much more slowly - confirm that the windows are raised.

### Submitter Checklist

### DIFF
--- a/src/bridges/labwc/labwc_bridge.py
+++ b/src/bridges/labwc/labwc_bridge.py
@@ -1148,6 +1148,7 @@ class Bridge:
         self.desktop_wm_preferences_changed(self.desktop_wm_preferences_settings, "titlebar-font")
         self.desktop_wm_preferences_changed(self.desktop_wm_preferences_settings, "button-layout")
         self.desktop_wm_preferences_changed(self.desktop_wm_preferences_settings, "num-workspaces")
+        self.desktop_wm_preferences_changed(self.desktop_wm_preferences_settings, "auto-raise-delay")
         self.desktop_interface_changed(self.desktop_interface_settings, "gtk-theme")
         self.desktop_interface_changed(self.desktop_interface_settings, "icon-theme")
         self.desktop_interface_changed(self.desktop_interface_settings, "cursor-theme")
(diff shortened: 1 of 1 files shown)


JoshStrobl (approved):
LGTM! Switched to mouse (tested with sloppy too) and can confirm there is now an expected delay. Thanks!

PR #939, opened 2026-06-22
panel: Modernize the panel plugin manager and fix lints

EbonJaeger:
## Description

Uses more modern practices, and generally cleans up some of the code. Fixed any lint issues while I was at it.

## Test plan

Installed and restarted the panel, see that all my applets loaded without exploding.

### Submitter Checklist

### DIFF
--- a/src/panel/plugin/plugin-manager.c
+++ b/src/panel/plugin/plugin-manager.c
@@ -11,12 +11,12 @@
 
 #include "plugin-manager.h"
 
-#include "budgie-config.h"
-#include "applet.h"
-#include "plugin.h"
-
 #include <girepository/girepository.h>
 #include <libpeas-2/libpeas.h>
+
+#include "applet.h"
+#include "budgie-config.h"
+#include "plugin.h"
 
 /**
  * BudgiePanelPluginManagerError:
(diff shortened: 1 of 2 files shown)


src/panel/plugin/plugin-manager.c
fossfreedom, line 416: hmm - in the create_applet at the bottom of the file

info = budgie_panel_plugin_manager_load_applet_instance(self, uuid, settings, NULL, &temp_err);

name is null - but here we are assigning the return value of g_strdup .... potential crash issue?

fossfreedom, line 426: maybe the same issue of assigning to a null value?

fossfreedom, line 447: ??? again potentially assigning against a null value?

src/panel/plugin/plugin-manager.c
```diff
@@ -54,17 +54,94 @@ struct _BudgiePanelPluginManager {
...
+static PeasPluginInfo *panel_plugin_manager_get_plugin_info(BudgiePanelPluginManager *self, const gchar *name) {
+	guint i = 0;
+	gpointer item;
+	PeasPluginInfo *info = NULL;
+	gchar *found_name = NULL;
+
+	while ((item = g_list_model_get_item(G_LIST_MODEL(self->engine), i)) != NULL) {
+		i++;
+
+		info = PEAS_PLUGIN_INFO(item);
```
fossfreedom: hmm - not sure on the macro here but I'm guessing its creating an object that we don't subsequently release on the second or further loop of the item - maybe something like g_clear_object(&info); ??

src/panel/plugin/plugin-manager.c
```diff
@@ -54,17 +54,94 @@ struct _BudgiePanelPluginManager {
...
+		found_name = peas_plugin_info_get_name(info);
```
fossfreedom, line 118: similar comment ?  does this allocate an object that needs releasing?

src/panel/plugin/plugin-manager.c
```diff
@@ -356,20 +366,20 @@ void budgie_panel_plugin_manager_modprobe(BudgiePanelPluginManager *self, const
...
+	g_return_val_if_fail(uuid != NULL, NULL);
+	g_return_val_if_fail(name == NULL || *name == NULL, NULL);
+	g_return_val_if_fail(err == NULL || *err == NULL, NULL);
+
 	// Make sure we have a valid settings instance for the plugin
-	path = create_applet_path(uuid);
+	path = panel_plugin_manager_create_applet_path(uuid);
 
 	if (plugin_settings == NULL) {
 		settings = g_settings_new_with_path(BUDGIE_APPLET_SCHEMA, g_strdup(path));
```
fossfreedom: do we actually need create another string g_strdup(path) ?  doesnt the hard work be done with the new "panel_plugin_manager_create_applet_path" ?

src/panel/plugin/plugin-manager.c
fossfreedom: maybe redundant to-do a g_strdup(plugin_name) ?

fossfreedom, line 451: should we validate the return value?  Can applet be invalid/null ?

fossfreedom (changes requested):

src/panel/plugin/plugin-manager.c
EbonJaeger: `peas_plugin_info_get_name`'s return is owned by the PeasPluginInfo, and we should not free it.

src/panel/plugin/plugin-manager.c
EbonJaeger: This isn't creating anything, it's casting to a pointer type (`PeasPluginInfo*`).

src/panel/plugin/plugin-manager.c
EbonJaeger: This is covered by the `g_return_val_if_fail(name == NULL || *name == NULL, NULL);` above.

EbonJaeger: Same as previous reply.

EbonJaeger: And again.

EbonJaeger: I think that if this returns something invalid, then there are bigger issues. I'm also not sure that this is in the scope of this PR's changes.

EbonJaeger: The code here *does* look a little strange, but this is how GObject C does optional out-arguments. Because we're assigning the *value* of `name`, not `name` itself, if that makes sense. And because of that, both `name` must be `NULL`, and the value pointed to by `name` must also be `NULL`.

src/panel/plugin/plugin-manager.c
EbonJaeger: That said, we should probably be freeing one of these at the end of the loop. Since they should both be pointing to the same memory, I don't know that it matters which one.

JoshStrobl (approved):
LGTM!

PR #941, opened 2026-06-27
panel/manager: Fix potential crash when checking for primary monitor

EbonJaeger:
## Description

One prevalent crash being caught by the Ubuntu error tracker is happening in the panel manager when we check if a window is on the primary monitor. Let's add some guards to hopefully prevent this crash.

## Test plan

Re-built Budgie, and restarted the panel.

### Submitter Checklist


JoshStrobl (approved):
LGTM, thanks!

PR #951, opened 2026-07-22
fix(panel): Super key doesn't focus app menu search box (#842)

### LINKED ISSUE #842: [Bug] Search field of Menu doesn't get input-focus if opened with Super_L
### Description

In version 10.10.1 it's not possible anymore to hit <Super_L> and start typing the name of the program to start. The input focus is simply not moved to the Menu.

Lets say I have a text editor open and hit Super_L .. the menu gets shown in the foreground, but the input-focus is still on my text editor. Now if I start typing a program name, all keyboard events are still sent to the text editor.

If I use the mouse to open the menu, the input-focus is correctly moved to the search-field and the cursor is blinking. But if I now type an app-name (e.g. "About") and select the item with the up-/down-keys, I can't hit the return-key to open the selected app.

### Budgie version

### Operating System

Arch Linux 6.18.7-arch1-1

1. Logon/Start a user session
2. Hit Super_L key
3. the Menu appears
4. the search field at the top isn't focused
5. type _something_ ... all keyboard input is still sent to the previous active window
6. Now even if I click at the search field (after opened with the keyboard) and type "About" and use the cursor-down to select the entry, I still can't start the program with the return-key

The following animation shows:
- first, how the menu acts if activated via mouse
- second, how the menu acts if activated via keyboard (here I'm unable to type something until it's activated by mouse)

Image

### Actual result

- If the menu is activated by keyboard, I can't directly start typing a search term .. window doesn't have input focus.
- An app selected via keyboard can't be opened via "enter/return" key

### Expected result

If the menu is activated by keyboard or mouse, I should be able start typing a search term .. 
If the menu is opened via keyboard and I click into the search-field, it should be possible to select an application-entry and hit enter to open it.

### Additional information

(I wasn't sure if I should've opened two issues: one for "missing focus", one for "unable to use return-key" ... give me a hint if I should do so. The enter-key-issue was just noticed while writing this down.)

avbuse:
## Description

Pressing `Super_L` opens the app menu popover visually, but keyboard input still goes to whatever window was previously focused. Opening the same menu with a mouse click works correctly.

This is fixable entirely in Budgie — it is **not** a compositor/wlroots bug and does **not** depend on labwc PR #3375 or wlroots MR !5265 (`xdg_popup` grab event). Those only affect `xdg_popup`-role surfaces; `GtkPopover` never uses that role under GTK3.

### Root cause (stacked)

1. **`GtkPopover` has no Wayland surface role of its own** — GTK3 renders it via a `wl_subsurface` and routes keyboard input through an in-process GTK grab. That grab only works if the parent toplevel (the panel) already holds real Wayland keyboard focus. The Super-key path never focuses the panel's `ON_DEMAND` layer-surface.
2. **`set_keyboard_interactivity(EXCLUSIVE)` is async** — focus must be grabbed from the toplevel's `focus-in-event` (GDK's representation of `wl_keyboard.enter`), not synchronously.
3. **Both call sites bypass `reset()`** — `invoke_action()` and the mouse handler call `manager.show_popover()` without `popover.show()`, so `BudgieMenuWindow.reset()` never marks the search entry as `focus_child`.
4. **`GtkPopover.grab_focus()` always does `TAB_FORWARD` descent** — even after `reset()`, a later `grab_focus(popover)` advances past the search entry onto the active category button.

### Fix

- `popover-manager.c`: toggle the panel layer-surface to `EXCLUSIVE` before showing, restore `ON_DEMAND` on hide; defer focus grab to `focus-in-event` (`G_CONNECT_AFTER`) when needed.
- `BudgieMenu.vala`: call `popover.reset(true)` before `show_popover()` at both call sites.
- `BudgieMenuWindow.vala`: override `grab_focus()` to target `search_entry` directly.

Fixes #842

## Test plan

- [x] Fresh clone, clean `ninja` build (499/499) on an independent checkout
- [x] Live verification: after pressing Super, typed characters land in the search box with zero clicks
- [x] Verified with `WAYLAND_DEBUG=1` protocol traces (popover is subsurface; panel gets EXCLUSIVE then reverts to ON_DEMAND)
- [ ] Mouse-click open still focuses search entry
- [ ] Esc / click-away / app launch restore panel keyboard mode (`ON_DEMAND`)
- [ ] Rapid double-Super does not leave stale focus handlers

### Submitter Checklist

Made with [Cursor](https://cursor.com)

### DIFF
--- a/src/panel/applets/budgie-menu/BudgieMenu.vala
+++ b/src/panel/applets/budgie-menu/BudgieMenu.vala
@@ -167,6 +167,12 @@ public class BudgieMenuApplet : Budgie.Applet {
 				popover.hide();
 			} else {
 				popover.get_child().show_all();
+				// Explicitly reset (which sets the search entry as the popover's
+				// tracked focus_child) before showing. manager.show_popover() does
+				// NOT call popover.show()/our BudgieMenuWindow.show() override, so
+				// without this the focus_child bookkeeping reset() performs never
+				// runs via this code path. See: https://github.com/BuddiesOfBudgie/budgie-desktop/issues/842
+				popover.reset(true);
 				this.manager.show_popover(widget);
 			}
 			return Gdk.EVENT_STOP;
@@ -216,6 +222,11 @@ public class BudgieMenuApplet : Budgie.Applet {
 				popover.hide();
 			} else {
 				popover.get_child().show_all();
+				// See the identical comment in the button_press_event handler above:
+				// manager.show_popover() never triggers our BudgieMenuWindow.show()
+				// override, so reset() (which sets the search entry as focus_child)
+				// must be called explicitly here too. See: https://github.com/BuddiesOfBudgie/budgie-desktop/issues/842
+				popover.reset(true);
 				this.manager.show_popover(widget);
 			}
 		}
--- a/src/panel/applets/budgie-menu/BudgieMenuWindow.vala
+++ b/src/panel/applets/budgie-menu/BudgieMenuWindow.vala
@@ -220,6 +220,28 @@ public class BudgieMenuWindow : Gtk.Popover {
 		base.show();
 	}
 
+	// GtkPopover hard-overrides grab_focus() (see gtk-3-24's gtkpopover.c,
+	// gtk_popover_grab_focus()) to unconditionally do:
+	//   child_focus(gtk_bin_get_child(popover), GTK_DIR_TAB_FORWARD)
+	// -- a relative TAB-order descent from whatever the popover's current
+	// focus_child already is. Since reset() already correctly sets focus_child to
+	// search_entry (via search_entry.grab_focus()), any later plain
+	// gtk_widget_grab_focus(popover) call (as done by popover-manager.c's deferred
+	// focus-in-event handler, needed to re-assert focus once real Wayland keyboard
+	// focus actually arrives) tabs PAST the already-focused search entry onto
+	// whatever's next -- landing on the active category radio button instead.
+	// Overriding grab_focus() here means popover-manager.c's existing
+	// gtk_widget_grab_focus(w) call (w = this popover) dispatches straight to
+	// search_entry instead, with no directional-descent side effects. This fixes
+	// both the initial reset() grab and the later re-assertion grab to agree on the
+	// same target. See: https://github.com/BuddiesOfBudgie/budgie-desktop/issues/842
+	public override void grab_focus() {
+		if (!this.get_mapped()) {
+			return;
+		}
+		this.search_entry.grab_focus();
+	}
+
 	/**
 	 * Opens our overlay menu and makes all other widgets insensitive.
 	 */
--- a/src/plugin/panel/popover-manager.c
+++ b/src/plugin/panel/popover-manager.c
@@ -31,6 +31,8 @@ G_DEFINE_TYPE_WITH_PRIVATE(BudgiePopoverManager, budgie_popover_manager, G_TYPE_
 
 static void budgie_popover_manager_widget_died(BudgiePopoverManager* self, GtkWidget* child);
 static gboolean on_focus_out(GtkWidget *widget, GdkEvent *event, GtkPopover *popover);
+static void on_popover_closed_restore_keyboard_mode(GtkPopover *popover, GtkWindow *toplevel);
+static gboolean on_toplevel_focus_in_grab_widget(GtkWidget *toplevel_widget, GdkEvent *event, GtkWidget *w);
 
 /**
  * budgie_popover_manager_new:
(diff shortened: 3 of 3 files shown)


avbuse:
# Root cause + fix for #842 — Super key doesn't focus app menu search box

## Summary

Pressing `Super_L` opens the app menu popover visually, but keyboard input
still goes to whatever window was previously focused. Opening the same menu
with a mouse click works correctly (focus lands in the search entry immediately).

This is **not** a compositor/wlroots bug and does **not** depend on
labwc PR #3375 or the draft wlroots MR !5265 (`xdg_popup` grab event) — those
only affect surfaces using the `xdg_popup` role, and `GtkPopover` never uses
that role under GTK3. This is fixable entirely in Budgie's own code.

Reproduced on: budgie-desktop 10.10.2, labwc 0.20.1, CachyOS (Arch), Lunar Lake laptop.

## Root cause (four independent, stacked bugs)

### 1. `GtkPopover` is a `GtkBin`, not a `GtkWindow` — it has no Wayland surface role of its own

GTK3 renders `GtkPopover` via a `wl_subsurface` parented to its toplevel
(the panel), and routes keyboard input through an **in-process GTK grab**,
not real Wayland popup/layer-shell semantics. Confirmed with a live
`WAYLAND_DEBUG=1 budgie-panel --replace` trace: across multiple popover
open/close cycles, exactly one `zwlr_layer_shell_v1.get_layer_surface()`
call ever appeared — for the panel bar itself, never for the popover. The
popover's surface is created via
`wl_subcompositor.get_subsurface(new_surface, panel_wl_surface)` every time.

This means the popover's in-process keyboard grab only has real key events
to redirect if its **parent toplevel** (the panel) already holds actual
Wayland `wl_keyboard` focus.

- **Mouse-click path works**: clicking the panel gives the panel's own
  `zwlr_layer_surface_v1` real keyboard focus (labwc honours the click per
  normal Wayland focus semantics), so the popover's internal grab has real
  input to intercept.
- **Super-key path fails**: the panel's layer-surface sits at
  `KEYBOARD_MODE_ON_DEMAND` (confirmed in the trace:
  `set_keyboard_interactivity(2)`). No click means no focus, means nothing
  for the popover's internal grab to redirect.

### 2. Fix location is the PANEL's existing layer-surface, and the toggle is asynchronous

The correct fix is to toggle keyboard-interactivity on the **panel's**
already-existing layer-surface to `EXCLUSIVE` right before opening the
popover via a keyboard shortcut, and revert to `ON_DEMAND` on close (every
hide path — Esc, click-away, app launch — routes through `hide`, so
connecting there covers all of them).

Critically, `set_keyboard_interactivity(EXCLUSIVE)` does **not** synchronously
hand over focus. `wl_keyboard.enter` arrives on a **later** Wayland
roundtrip. A first attempt that called `gtk_widget_grab_focus()` in the same
function call as the EXCLUSIVE request silently failed — it grabbed GTK
focus on a toplevel that didn't have real Wayland focus yet. The fix:
connect to the toplevel `GtkWindow`'s own `"focus-in-event"` signal (GDK's
Wayland backend translates a `wl_keyboard.enter` into exactly this — no
finer-grained public GTK3 signal exists for it) and grab focus from inside
that handler, one-shot, using `g_signal_connect_object(..., G_CONNECT_AFTER)`
(the `_AFTER` matters: GtkWindow's own default focus-in handler
re-establishes toplevel-focus bookkeeping against its *stored* focus widget
when real focus arrives — running before that can get clobbered).

### 3. Both call sites (mouse AND keyboard) bypass `reset()` entirely

`invoke_action()` (Super-key path) and the mouse-click handler both call
`this.manager.show_popover(widget)` directly — **neither ever calls
`popover.show()`**, which is the only thing that invokes
`BudgieMenuWindow`'s `show()` override, which is what calls `reset()` (the
method that sets the search entry as the popover's tracked `focus_child` via
`search_entry.grab_focus()`). Fix: call `popover.reset(true)` explicitly
before `manager.show_popover(widget)` at both call sites.

### 4. `GtkPopover.grab_focus()` is hard-overridden to do a directional TAB descent — verified against GTK3 source

Even after fixes 1–3, the deferred grab still landed on a random category
button instead of the search entry. GTK3's own source (`gtkpopover.c`)
shows why:

```c
widget_class->grab_focus = gtk_popover_grab_focus;   // line 1726

static void
gtk_popover_grab_focus (GtkWidget *widget)            // line 1616
{
  GtkPopoverPrivate *priv = GTK_POPOVER (widget)->priv;
  GtkWidget *child;
  if (!priv->visible)
    return;
  child = gtk_bin_get_child (GTK_BIN (widget));
  if (child)
    gtk_widget_child_focus (child, GTK_DIR_TAB_FORWARD);   // <-- always this
}
```

`gtk_widget_grab_focus(popover)` is **hard-coded** to do
`child_focus(bin_child, GTK_DIR_TAB_FORWARD)` — a relative descent from
whatever the popover's current `focus_child` already is. Since `reset()`
correctly set `focus_child` to the search entry, the directional descent
advances **one step past it**, landing on the next focusable thing in tab
order (the active category radio button). `set_active(true)` on that button
does *not* independently steal focus as a side effect — `GtkToggleButton`/
`GtkRadioButton` only grab focus on real button-press, gated by
`focus-on-click`. The steal is purely the directional-descent behavior.

**Fix**: since a generic popover-manager holding only the container widget
handle cannot force focus onto one specific descendant (any `grab_focus()`
call on the container is either a no-op or this hard-coded descent),
override `grab_focus()` in the `BudgieMenuWindow : Gtk.Popover` subclass
itself, which knows its own internals:

```vala
public override void grab_focus() {
    if (!this.get_mapped()) {
        return;
    }
    this.search_entry.grab_focus();
}
```

Now the popover-manager's existing `gtk_widget_grab_focus(w)` call (where
`w` is the popover) dispatches straight to the search entry — a real
last-writer-wins grab, immune to tab order and stale `focus_child` state.
No change needed to the generic manager API.

One thing worth flagging: `GtkPopover`'s modal-focus path
(`gtk_popover_apply_modality`) also calls `grab_focus(popover)` — this
override correctly redirects that too, which is harmless/correct for a
search-first menu. In-menu keyboard *navigation* (arrow keys/Tab while the
menu is open) goes through the separate `focus` vfunc
(`gtk_popover_focus`), not `grab_focus`, so this override does not affect
normal menu navigation.

## Verified fix (all four pieces)

Confirmed live: typed characters land in the search box with zero clicks,
immediately after pressing Super. No labwc/wlroots changes needed at all.

### `src/plugin/panel/popover-manager.c` (in `budgie_popover_manager_show_popover`)

```c
GtkWindow *toplevel_win = GTK_IS_WINDOW(toplevel) ? GTK_WINDOW(toplevel) : NULL;

if (toplevel_win != NULL && gtk_layer_is_layer_window(toplevel_win)) {
	gtk_layer_set_keyboard_mode(toplevel_win, GTK_LAYER_SHELL_KEYBOARD_MODE_EXCLUSIVE);

	g_signal_connect(popover, "hide", G_CALLBACK(on_popover_closed_restore_keyboard_mode), toplevel_win);

	if (gtk_window_has_toplevel_focus(toplevel_win)) {
		// Panel already had real keyboard focus (e.g. clicked moments before
		// Super was pressed) -- EXCLUSIVE produces no NEW wl_keyboard.enter,
		// so waiting for focus-in-event would hang forever. Grab immediately.
		gtk_widget_grab_focus(w);
	} else {
		g_signal_handlers_disconnect_by_func(toplevel_win, G_CALLBACK(on_toplevel_focus_in_grab_widget), w);
		g_signal_connect_object(toplevel_win, "focus-in-event", G_CALLBACK(on_toplevel_focus_in_grab_widget), w, G_CONNECT_AFTER);
	}
} else {
	gtk_widget_grab_focus(w);
}
```

```c
static gboolean on_toplevel_focus_in_grab_widget(GtkWidget *toplevel_widget, GdkEvent *event, GtkWidget *w) {
	if (GTK_IS_WIDGET(w) && gtk_widget_get_visible(w)) {
		gtk_widget_grab_focus(w);
	}
	g_signal_handlers_disconnect_by_func(toplevel_widget, G_CALLBACK(on_toplevel_focus_in_grab_widget), w);
	return GDK_EVENT_PROPAGATE;
}

static void on_popover_closed_restore_keyboard_mode(GtkPopover *popover, GtkWindow *toplevel) {
	if (toplevel != NULL && GTK_IS_WINDOW(toplevel) && gtk_layer_is_layer_window(toplevel)) {
		gtk_layer_set_keyboard_mode(toplevel, GTK_LAYER_SHELL_KEYBOARD_MODE_ON_DEMAND);
	}
	g_signal_handlers_disconnect_by_func(popover, G_CALLBACK(on_popover_closed_restore_keyboard_mode), toplevel);
	g_signal_handlers_disconnect_by_func(toplevel, G_CALLBACK(on_toplevel_focus_in_grab_widget), popover);
}
```

### `src/panel/applets/budgie-menu/BudgieMenu.vala`

In both the mouse `button_press_event` handler and `invoke_action()`, add
before `this.manager.show_popover(widget)`:

```vala
popover.reset(true);
```

### `src/panel/applets/budgie-menu/BudgieMenuWindow.vala`

```vala
public override void grab_focus() {
	if (!this.get_mapped()) {
		return;
	}
	this.search_entry.grab_focus();
}
```

## Notes for reviewers

- This is fully independent of labwc PR #3375 / wlroots MR !5265 — please
  don't gate this fix on those landing. They address `xdg_popup`-role
  surfaces specifically; `GtkPopover` never has that role under GTK3.
- Tested via live `WAYLAND_DEBUG=1` protocol traces at every step — happy to
  share the full traces if useful for review.
- The `EXCLUSIVE`→`ON_DEMAND` toggle is scoped tightly to the popover's
  open/close lifecycle so the panel doesn't trap the compositor keyboard
  globally between opens.

avbuse:
retracted to complete testing plan.

JoshStrobl:
I appreciate this but I already have panel and popover management overhaul code in the works based off my learnings from an internal test of GTK4 porting (that won't see the light of day but many architectural bits will), so I'm going to close this in preference of that work.

JoshStrobl:
Also I am curious as to why you didn't stop and question why there was countless lines of comments (more comments than code changes isn't good) and curtail your agent from doing even more, nor take the time to clean them up. Please be more proactive with your agent. You are responsible for this code. There is no way in hell I'm going to spend the time reading all those comments, nor the wall of text from the PR.

P.S. We don't use Co-Authored by for agents, only the assisted-by.

PR #955, opened 2026-08-25
Support dynamically defined media-keys

fossfreedom:
## Description

This PR supports predefined dynamic keybinds; what I mean by this are media-keys that should perform a specific action if the user/distro has a specific app installed and the user has set the media-key keybind via budgie-control-center - keyboard

So the example here is for users that have autostarted wvkbd - the wlroots compatible on-screen keyboard

So I could envisage we could support other options that distros want to set e.g. squeekboard and other on-screen keyboards. Some distros I believe use wvkbd-desktopxxxx or something like that - on debian/ubuntu we use wvkbd-mobintl as the add executable.

Likewise we could add in the future other apps for media keys defined in rc.xml that are bridgeable.

I've kept this simple for this PR - everything is defined in an array in the bridge.  We could extend this to be truly user/distro definable via a separate file to replace the array - but thats a future enhancement.

## Test plan
1. On Ubuntu Budgie 26.04/26.10 install wvkbd
2. run wvkbd via `wvkbd-mobintl -l full --hidden`
3. Set a keybind C-A-k for the media-key on-screen-keyboard - note for UB 26.10 we'll be enabling this via BCC by enabling (true) in the BCC labwc keyfile
4. Look at rc.xml - note it has defined an action and command for wkvd-mobintl
5. Clear the keybind in BCC - look at rc.xml and the action & command has been removed.

### Submitter Checklist


JoshStrobl (approved):
I don't have this tool available under Fedora, but the code itself looks good and the shutils.which check for the executable is sufficient in not introducing any issues in my testing, so approving. Thanks! Might have to see about getting this packaged for Fedora.

Issue #956, opened 2026-08-26
[Bug] Labwc shows error on logon


Namelus11811:
### Description

When logging into the desktop labwc displays an error since labwc version 0.20.2

### Budgie version

budgie-desktop 10.10.2 (git-119c4aceab8f45f6adcd9be4ed1fb542e3a40b85)

### Operating System

postmarketOs Edge

### Steps to reproduce the issue

Logon to budgie-desktop and observe error window, from labwc

### Actual result

Error window created by labnag on logon
<img width="1366" height="768" alt="Image" src="https://github.com/user-attachments/assets/0fb29af4-dff5-4eac-b06f-26d86e664432" />

### Expected result

No error windows on login

### Additional information

Labnag error message was added with latest labwc version 0.20.2
These errors in config file always existed, and are created by the labwc bridge putting empty keybinds into the config file. I would suggest either disabling the keybind, or changing the emty keybind behaviour

fossfreedom:
Noted - thx for raising.

We saw changes were afoot a while back - looks like labwc have pushed the proverbial button

https://github.com/BuddiesOfBudgie/budgie-desktop/issues/926

fossfreedom:
https://labwc.github.io/labwc-config.5.html

For an immediate workaround worth playing around adding errorcommand as per the above doc into your rc.xml - might work if you leave an empty value or add some-sort of no-op type of operation as the suggested command.

xcorvis:
Just noting I'm also seeing this on CachyOS with budgie-desktop 10.10.2-2.1and labwc 0.20.2-1.1.

From the sound of it this can be safely dismissed and ignored, right?

fugo:
Arch Linux Budgie-Desktop 10.10.02
labwc 0.20.2 (+xwayland +nls +rsvg +libsfdo) wlroots-0.20.2

since the error display it also changed my command:
`<keybind bridge="custom5" key="A-KP_6">` to `key="A-6"`

Hi-Phile:
This also appears in the Budgie Desktop Settings Panel when accessing or changing the Labwc Compositor Theme from the pull-down menu.

PR #962, opened 2026-09-07
feat(bridges): support dynamic keybinds handling

fossfreedom:
## Description

Move all keybinds into a separate shippable file keybind.xml.example.  Distros can override this to 
customise the keybinds they wish to ship with - in the same data folder call it keybind.xml

Bridge updated extensively to both cleanup existing rc.xml keybind entries (hence fixing #956) and
now to merge in keybinds that is syntactically correct.  

Keybinds can now support "optional" executables - so if those executables are not installed
then the keybind is not merged.

## Test plan
using labwc 0.20.1 and previous
0. before upgrading make a copy of your local rc.xml and keep it safe. Upgrade to the keybinds branch. 
1. New user - no ~/.config/budgie-desktop folder - login and examine rc.xml and related files are created from the keybind.xml.example file
2. Existing user - examine ~/.config/budgie-desktop folder - examine rc.xml and look for empty action attributes and undefined keys - they should have all been removed
3. Create custom keys - test that they work - look in rc.xml to confirm
4. Remove custom keys - test that they have been removed - look in rc.xml to confirm
5. Define a keybind available in BCC that was previously set to disabled - test that works and confirm in rc.xml that any entry exists
6. Remove/undefine a keybind available in BCC - test that the keybind no longer works and confirm in rc.xml that any entry exists
7. Static keys - manually define in rc.xml your own keybinding - Use action Execute and a command to show that it runs your command
8. Static keys - check in your local a static keybind and confirm that it works.
9. killall labwc_bridge.py --> i.e. kill the bridge
10. copy back your local backup of rc.xml over the top of your copy current rc.xml
11. logout and login
12. look at the local rc.xml - see the new attributes managed etc.
13. confirm that static keybinds are working
14. confirm that your locally created keybind is not marked as managed - i.e. hasn't been deleted or modified - kept as is.

Now repeat all of the above with labwc 0.20.2 and later

You will note that labwc 0.20.2 will throw a labnag error message on initial login - this is because at initial login the older - now invalid - rc.xml is there. The bridge will the start and the rc.xml will be cleaned up causing the red labnag dialog to disappear. 

### Submitter Checklist

- [X] If AI tools or LLMs were used as part of this contribution, I have read the [AI Policy](https://docs.buddiesofbudgie.org/organization/ai-policy) and commits include `Assisted-by` trailers where applicable


src/bridges/labwc/rc.xml
```diff
@@ -217,87 +24,7 @@
 		<keybind key="XF86AudioStop">
 			<action name="Execute" command="dbus-send --type=method_call --dest=org.buddiesofbudgie.BudgieMprisController /org/buddiesofbudgie/MprisController org.buddiesofbudgie.BudgieMprisController.Stop" />
 		</keybind>
```
JoshStrobl, line 26: Was there any specific line of thinking on having these in the rc.xml and not the keybinds.xml.example?

fossfreedom: Reworked to include static keybind handling - see the test plan

EbonJaeger (approved):

JoshStrobl (approved):
Thanks for the all the work! I did some cleanup of existing code to fix some issues raised by tuff and basedpyright, as well as made the systemd journal import optional.

Stronger runtime dep check in meson too.

Will do a broader restructure to transform the bridge code into multiple files to be easier to mentally parse, but this'll land as-is and that will be a subsequent branch.

PR #963, opened 2026-09-08
fix(session) switch bash for /bin/sh in startbudgielabwc script

Namelus11811:
## Description

In startbudgielabwc script `#!/usr/bin/env bash` has been replaced with `/bin/sh`.  This makes the script POSIX compatible, and removes the need for bash 

## Test plan

Edited /usr/bin/startbudgielabwc with change on running system, uninstalled bash, and rebooted. Logged in with no issues.
Tested on busybox ash and dash

### Submitter Checklist

Small commit, ties into bigger commit on budgie-session repository

### DIFF
--- a/src/session/startbudgielabwc.in
+++ b/src/session/startbudgielabwc.in
@@ -1,4 +1,4 @@
-#!/usr/bin/env bash
+#!/bin/sh
 OPTS=""
 for OPT in "$@"; do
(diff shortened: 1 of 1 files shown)


JoshStrobl (approved):
LGTM! Will get this tested after work and land assuming I get into a session 😁 (I almost certainly will).

Thanks for your contribution!

PR #964, opened 2026-09-08
fix(budgie-menu): focus search entry when opened via keybinding

### LINKED ISSUE #842: [Bug] Search field of Menu doesn't get input-focus if opened with Super_L

JoshStrobl:
## Description

During the move to Wayland, our Budgie Popovers were converted from being Gtk.Window to GtkPopover. On Wayland, popovers are considered a subsurface of its parent (in our case the panel), which only gets keyboard input while the panel itself is focused, and a keybinding can't focus the panel.

Super_L showed the menu but typing went to the previous window, so you'd have to click into the search input to start typing.

Made the menu its own toplevel layer-shell window again, this time with some custom positioning against the launcher.

I have tested it against all panel positions as well as doc mode. Some code churn from moving some of the functions away from being lambdas.

Closes #842

Assisted-by: Claude:claude-fable-5-1

## Test plan

- Install branch
- Restart budgie panel
- Open Budgie Menu using Super_L
- Expected result is that you should now have input in the search

### Submitter Checklist


src/panel/applets/budgie-menu/BudgieMenuWindow.vala
```diff
@@ -215,9 +298,163 @@ public class BudgieMenuWindow : Gtk.Popover {
...
+		bool vertical = (position == Budgie.PanelPosition.LEFT || position == Budgie.PanelPosition.RIGHT);
+
+		int monitor_extent = 0;
+		if (monitor != null) {
+			Gdk.Rectangle geometry = monitor.get_geometry();
+			monitor_extent = vertical ? geometry.height : geometry.width;
+		}
+
+		// Where the launcher sits along the panel and how big it is. The panel
+		// starts at the monitor edge, so panel coordinates are monitor coordinates.
```
EbonJaeger, line 350: I don't think this assumption is always true. What happens if the panel is in dock mode?

EbonJaeger: I guess this is addressed a little further on in the code.

src/theme/3.20/_main.scss
```diff
@@ -31,6 +31,19 @@
     "undershoot",
     "overshoot"
   );
+
+  border-radius: 3px;
```
EbonJaeger, line 35: What's this from? Do other applets have that?

JoshStrobl: No it's just to try to imitate the radiusing of popovers.

src/panel/applets/budgie-menu/BudgieMenuWindow.vala
JoshStrobl: Correct. For example, a horizontal panel would have us getting the allocated width, subtracting monitor width from it to get remaining space, then dividing by 2 for the left/right offsets. Starts on L363.

EbonJaeger (approved):
LGTM, thanks!

PR #965, opened 2026-09-10
fix(raven): anchor to the vertical edges instead of sizing from panels

### LINKED ISSUE #843: Budgie 10.10.2: Raven side-panel overlaps the Panel
Budgie 10.10.1 on Arch.

Raven side-panel overlaps the Panel when display scaling is set 2x.
(Panel is at the bottom of the screen)

JoshStrobl:
## Description

Raven requested a height of the screen minus the panel sizes while being anchored to one side edge only, so the compositor centered it in the usable area. Whenever those two disagreed it hung over both ends, which is visible as Raven overlapping a bottom panel. Anchoring to TOP and BOTTOM now leaves the height calculation to the compositor, which derives the usable area from all the exclusion zones.

Corrected some panel calculations that accidentally resulted in the "main layout" / content being eaten up by the shadow size. It was only 5 pixels, but this now will actually treat the size set by the user as the actual size of the content, not content - shadow. This "reserved size" is implemented in the `Toplevel` since that is what the panel manager expects as the interface for all the panels.

Also gutted the panel's remaining scale factor usage. The animation buffer no longer multiplies its size by the scale factor. History tells us having scaling code is bad when GTK is handling it, so let's rip this out.

Closes #843

## Test plan

1. Used Budgie Desktop before patch, opened wdisplays and change primary monitor (that the panel lives on) to 2.0 scaling. Opened up Raven and confirmed overlap. Change scaling back to 1.0 so I could actually use my computer.
2. Installed budgie-desktop with fix.
3. Replaced panel (`budgie-panel --replace &`).
4. Re-set scaling back to 2.0, opened up Raven and noted it no longer overlaps because the height setting is handled by the compositor.

Screenshot below from 2x scaling:

raven-scale-no-overlap

### Submitter Checklist

### DIFF
--- a/src/lib/toplevel.vala
+++ b/src/lib/toplevel.vala
@@ -37,6 +37,22 @@ namespace Budgie {
 		public int intended_size { public set ; public get; }
 
 		/**
+		* The extent of our window along the same axis as intended_size,
+		* including the shadow when it is visible
+		*/
+		public int targeted_size {
+			public get {
+				return intended_size + (shadow_visible ? ShadowBlock.SIZE : 0);
+			}
+		}
+
+		/**
+		* Thickness of the panel itself as allocated, excluding the shadow.
+		* This is the space we take from other surfaces
+		*/
+		public int reserved_size { public set; public get; }
+
+		/**
 		* Our configured applet spacing
 		*/
 		public int intended_spacing { public set; public get; }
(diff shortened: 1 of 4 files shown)


EbonJaeger (approved):
LGTM, thanks!

PR #969, opened 2026-09-11
fix(budgie-menu): hide the menu, not the toplevel, when a drag ends

### LINKED ISSUE #929: [Bug] Bottom panel crash after drag & drop to desktop
### Description

When you drag and drop any program icon from the Budgie menu to the desktop the bottom panel disapears. 
I could only manage to bring it back by opening Budgie Desktop Settings (desktop right-click) > Bottom Panel and switch Position to Top and then back again to Bottom.

BTW: `budgie-session --version` doesn't show any output on my system, hence I can only provide the Fedora package version below.

### Budgie version

F44 package-version: 10.10.2-1.fc44

### Operating System

Fedora 44 (kernel 7.0.10-201.fc44.x86_64)

1. Click Budgie menu
2. Drag any application icon to the desktop
3. Bottom panel disappears

### Actual result

Bottom panel invisible, possibly crashed.

### Expected result

Obviously the bottom panel should not be affected by drag & drop to desktop.

### Additional information

This system is not a fresh installation but the result of a standard Fedora system-upgrade 43 -> 44, whick performed without errors or warnings. Budgie desktop was already installed - and infact working very well - before the upgrade.

JoshStrobl:
## Description

MenuButton hid its toplevel once a drag finished. Until 0e3ab119 the menu was a Gtk.Popover, and a widget inside a popover reports the enclosing window as its toplevel, so dragging an app icon onto the desktop hid the panel. Switching the panel position and back re-showed it, which is the workaround in the issue.

The menu is its own Gtk.Window now, which is why this stopped happening on main. Walk up to the BudgieMenuWindow ancestor instead, so we hide the window the button actually belongs to.

Closes #929

## Test plan

There is no behavioral change from this commit, it's simply changing the logic of what we are hiding to be more explicit about finding a BudgieMenuWindow. That said, you can test by:

1. Installing the patch
2. Drag an app (non-flatpak unless you have latest commit in budgie-desktop-view) to the desktop, assuming desktop icons enabled
3. Menu closes correctly

### Submitter Checklist

### DIFF
--- a/src/panel/applets/budgie-menu/BudgieMenuButtons.vala
+++ b/src/panel/applets/budgie-menu/BudgieMenuButtons.vala
@@ -99,8 +99,12 @@ public class MenuButton : Gtk.Button {
 		);
 	}
 
-	private bool hide_toplevel() {
-		this.get_toplevel().hide();
+	private bool hide_menu() {
+		unowned var menu = this.get_ancestor(typeof(BudgieMenuWindow)) as BudgieMenuWindow;
+		if (menu != null) {
+			menu.hide();
+		}
+
 		return false;
 	}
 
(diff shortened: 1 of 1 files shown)


EbonJaeger (approved):
LGTM, thanks!

PR #970, opened 2026-09-11
fix(panel): keep settings in sync after a panel is moved

### LINKED ISSUE #832: Panel changes cause a variety of minor issues
Some initial testing:

image

Top panel on login - move the panel right and the label on the left doesnt get updated until you close & open settings.

After than the create panel button doesn't do anything

Maximise a window such as firefox on the right and there is a couple of pixel separation - not theme dependent

image

So intellihide hides the dock in maximised - good. Autohide doesn't do anything on maximise though i.e. still behaves like a static panel

As you have noted - difficult to get the panel to reappear - workaround would be to press Super to open the menu. But closing the menu doesnt make the panel hide again.

_Originally posted by @fossfreedom in https://github.com/BuddiesOfBudgie/budgie-desktop/issues/829#issuecomment-3828540681_

JoshStrobl:
## Description

Moving a panel destroys it and recreates it under a new object, but the settings page kept pointing at the closed one, so its position notify never fired again. Emit panel_added for the recreated panel and rebind the existing page to it instead of returning early on a UUID we already have. The applets page is built against a single panel, so it gets replaced, and it now disconnects from the panel it was built with on destroy.

rebuild_position_model cleared the model without restoring the selection, which is what left the combobox blank. Adding or removing another panel did the same through panels_changed.

Moving also left the edge the panel came from marked as in use, since show_panel applies the new position before set_placement can release it. Once all four edges were claimed there were no slots left and the create panel button did nothing.

Here is a video of it now functioning as expected: https://youtu.be/sgwvVdAULzY

Closes #832

## Test plan

1. Prior to use of patch, open up Budgie Desktop Settings and change panel position.
2. Notice that the dropdown gets cleared. You have to restart Budgie Desktop Settings for it to update (but even then you aren't able to create panels against all slots.
3. Install patch and restart budgie panel: `budgie-panel --replace &`
4. Re-open Budgie Desktop Settings, change panel positions, notice it updates immediately

### Submitter Checklist

### DIFF
--- a/src/panel/manager.vala
+++ b/src/panel/manager.vala
@@ -1393,6 +1393,36 @@ namespace Budgie {
 
 			// Load panel again
 			load_panel(uuid, true);
+
+			// Moving leaves the old edge marked as in use, since show_panel() applies
+			// the new position before set_placement() can release it. Left alone,
+			// get_first_position() runs out of edges and new panels stop being created
+			this.update_slots();
+
+			// Anything holding the panel we just closed has to pick up the new one
+			this.panel_added(uuid, panels.lookup(uuid));
+		}
+
+		/**
+		* Recalculate which screen edges are in use from the panels we have now
+		*/
+		void update_slots() {
+			unowned Screen? area = screens.lookup(primary_monitor);
+			if (area == null) {
+				return;
+			}
+
+			string? key = null;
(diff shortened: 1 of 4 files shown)


EbonJaeger (approved):
LGTM, thanks!

PR #971, opened 2026-09-11
feat(budgie-menu): add application favorites

JoshStrobl:
## Description

Right-clicking a menu item reveals an add/remove favorites entry and the application's desktop actions. Favorited applications get added to a new Favorites category above All, which appears once there is something in it. Compact mode has no category list, so it sorts favorites to the top of its single list instead.

The desktop action list moved out of icon-tasklist into Budgie.ApplicationActionList, and the parsing behind it onto Budgie.Application, which already holds the DesktopAppInfo that ButtonPopover was re-creating to resolve action names.

CategoryButton and MenuButton moved out of BudgieMenuButtons.vala into buttons/CategoryButton.vala and buttons/MenuButton.vala, one class per file, for a clearer separation of concerns. Not the biggest fan of multiple classes per file (unless they're small abstract classes). 

Also drops PropertiesRemote's PropertiesChanged declaration. It took no arguments, so GDBus discarded every incoming PropertiesChanged and logged a signature warning, once per AccountsService property update.

## Test plan

1. Run budgie-desktop from basically any point in the last 50 years
2. Notice no ability to favorite an app
3. Install this patch and restart budgie-panel (`budgie-panel --replace &`)
4. Right click an app entry in Budgie Menu
5. See new revealer, click "Add to Favorites" button
6. See new Favorites category
7. Close Budgie Menu and re-open it
8. Notice Favorites is then the default
9. Search still works even when on favorites, will just search everything

### Submitter Checklist


src/panel/applets/budgie-menu/FavoritesManager.vala
```diff
@@ -0,0 +1,72 @@
...
+	private string[] favorites;
+
+	public FavoritesManager(Settings settings) {
+		Object(settings: settings);
+	}
+
+	construct {
+		this.reload();
+
+		this.settings.changed["favorites"].connect(this.on_settings_changed);
```
EbonJaeger: Should these be swapped? If I understand correctly, we won't get the updates via the signal until the property is first read once the signal handler has been connected.

src/panel/applets/budgie-menu/FavoritesManager.vala
```diff
@@ -0,0 +1,72 @@
...
+
+		foreach (unowned var id in this.favorites) {
+			if (id != desktop_id) {
+				updated += id;
+			}
+		}
+
+		if (!this.is_favorite(desktop_id)) {
+			updated += desktop_id;
+		}
```
EbonJaeger: I think this could be better if we don't rebuild the list unless we actually need to. If an application is being added as a favorite, we only need to add to the array, and there is no need to do a full iteration and reconstruction.

JoshStrobl: That's fair. I originally had the implementation set as a GenericSet so it'd be a simple push, but then you have to reconvert back to a string array when you want to set it into settings, so I ended up changing it to a string[] through-and-through.

In reality, neither are going to be particularly harmful to performance, but I'll swap it back to a GenericSet :+1:

src/panel/applets/budgie-menu/FavoritesManager.vala
JoshStrobl: that'd be really dumb of gobject but i won't question it and will swap them kekw

src/panel/applets/budgie-menu/FavoritesManager.vala
EbonJaeger: I don't think you have to swap to a Set or List. Couldn't you check if the app is favored, and if not, add it to the array, and if not, reconstruct?

JoshStrobl: big true. should be fixed in 39e74dc85e4d0dc459c1a41289160f95b6b7c870

src/panel/applets/budgie-menu/FavoritesManager.vala
JoshStrobl: fixed in 39e74dc85e4d0dc459c1a41289160f95b6b7c870

EbonJaeger: Yeah, it really is. I don't know what the reason is; maybe performance?

The docs just say:
> Note that `settings` only emits this signal if you have read key at least once while a signal handler was already connected for key.

EbonJaeger (approved):
LGTM, thanks!

PR #972, opened 2026-09-12
fix(raven): handle toggles that arrive while closing or animating

JoshStrobl:
## Description

Clicking a panel applet moves keyboard focus to the panel, so Raven's focus-out closed it before the applet's toggle request arrived over DBus. The toggle then found Raven closed and opened it again. Ignore a toggle that lands within 250ms of a focus-out close, so it counts as that close instead of reopening.

Resolves a regression introduced with the fix in 016f53d411a3cd0b7d2fd381ad89b1d056c83fef for Raven not closing on focus out.

Rapid toggling left state behind as well. set_expanded built a new Animation on every call without stopping the previous one, so both wrote nscale each frame, and the close path queued an untracked timeout that hid the window 100ms later, which fired even if Raven had reopened by then. Store the animation and the timeout id as fields, then stop the animation and remove the timeout at the start of set_expanded. The slide now starts from the current nscale so a reversal does not snap.

## Test plan

1. Use budgie-desktop against main
2. Open Raven using applet
3. Attempt to close Raven using applet, notice it re-opens
4. Spam open/close via keyboard shortcut, notice animation jank and inconsistency behavior
5. Apply patch
6. Both issues fixed :)

Tested against Click and Mouse focus methods

### Submitter Checklist


src/raven/raven.vala
```diff
@@ -97,41 +97,58 @@ namespace Budgie {
...
+		private bool should_expand() {
+			if (this.is_expanded) {
+				this.is_expanded = false;
+				return false;
+			}
+
+			// If we are or just focused out of the window, Raven is currently closing
+			// So we do this check to not immediately re-trigger Raven to open
+			return !parent.consume_focus_out_close();
+		}
```
EbonJaeger, line 113: This function seems weird to me. The name implies that it only returns whether or not Raven should expand. It also closing Raven if it is already open feels like a strange side-effect.

EbonJaeger: I haven't been able to come up with a better or cleaner solution. This might be a case of "don't let perfect be the enemy of good", and keep this change as-is.

JoshStrobl: Fair enough. In that case, I'll merge the changes in. It can always be improved down the road :)

PR #975, opened 2026-09-13
chore: purge remains of libwnck

JoshStrobl:
## Description

We do not use libwnck directly anymore. we don't support x11 and therefore don't need any query x11 atom, or wnck window querying code.

Any window querying is expected to be done via libxfce4windowing, which itself may be able to be wayland-only (at compilation time) for distributions that shall not support X11.

As a result, purged all the old AppSystem code (been replaced with Windowing anyways), which was otherwise unused in the tab switcher (itself unused but will clean up in a separate PR).

## Test plan

Given nothing that we use in the Wayland session actually uses it, tested beyond compilation is pretty pointless. Regardless, I did the following:

1. Compiled budgie-desktop with patch
2. Restarted budgie-panel and budgie-daemon (both with `--replace &)
3. Confirmed that nothing exploded

### Submitter Checklist


EbonJaeger (approved):
LGTM, thanks!

PR #976, opened 2026-09-15
fix(budgie-menu): make budgie-menu visibie while fullscreen-mode is used

jls17:
## Description

I did a test of the changes from bdd3251 / #964 .

There is a small issue with the chosen layer of the new top-level window. Right now its set to "TOP" which places it next to all other desktop windows. Now if you use some kind of video player in fullscreen mode, hitting Super_L brings the budgie-menu up behind the current fullscreen app, thus its invisible to the user. Budgie-menu even gets the input focus and if you blindly type an application name and hit enter, you could start something. But previously the used `GtkPopover`-windows where visible above a fullscreen window. That's why I think the layer of the  budgie-menu should be `OVERLAY`.

## Test plan

- use a browser to playback a video (alternative: use VLC)
- activate fullscreen mode in the video player
- hit Super_L (this brings up budgie-menu in the background)
- type the name of some application (e.g. "term) and hit enter
- the app launches and is shown

After the layer fix, budgie-menu is visible if you hit Super_L

### Submitter Checklist

- [X] If AI tools or LLMs were used as part of this contribution, I have read the [AI Policy](https://docs.buddiesofbudgie.org/organization/ai-policy) and commits include `Assisted-by` trailers where applicable


JoshStrobl (approved):
Thanks for the fix! Still more work to be done on this front related to panels (out of scope of this work), but this was a clear regression from my change. Good catch 👍

PR #977, opened 2026-09-16
chore(session): enable specifying budgie-session libexecdir and gsd components

JoshStrobl:
## Description

**budgie-session libexecdir option:**

budgie-desktop looked for budgie-session-compositor-ready under gsd-libexecdir, which only works when budgie-session and gnome-settings-daemon install their executables to the same place.

The new option lets distributions point at budgie-session directly and defaults to the same libexecdir as before.

Both fallbacks now use libexecdirroot, which is the same path the gsd one was spelling out inline.

**gnome-settings-daemon component (inclusion / exclusion) options:**

Not all distributions support gsd-usb-protection and in the case of disk-utility-notify, that requires gnome-disks be installed on the system.

In lieu of being able to fire up these components conditionally if the executable exists, this option will enable distributions to curate the GSD components available. This comes with the understanding that not all capabilities will necessarily work out-of-the-box if excluded, as the desktop entry wouldn't be installed and thus the gsd-* not ran.

There is also a gsd-components-excluded option, which is subtracted from gsd-components. If the list of components you need to exclude is considerably shorter than the total amount, say you only need to exclude one, you can use that instead of listing every component you want to keep.

This should fix some journal noise from systemd-xdg-autostart-generator for when exec binaries are empty.

## Test plan

Not so much a reproducible test plan but rather _how_ I tested it

1. Worked through packaging budgie-desktop stack for aerynOS
2. No support for DiskUtilityNotify or UsbProtection, so needed to be able to turn that off to save the systemd-xdg-autogenerator noise. Opted to support both inclusion and exclusions, so distributions that have few options can use the inclusion, and those with most-but-not-all can roll with exclusion.
3. Noticed budgie-session execs like the compositor check were getting put into `/usr/lib/budgie-session/*`, but our meson was forcing it to then be the same as the gsd-libexecdir which was `/usr/lib/gnome-settings-daemon/`. For obvious reasons, budgie-session execs aren't pushed there.

### Submitter Checklist


EbonJaeger (approved):
LGTM, thanks!

PR #978, opened 2026-09-17
feat(keyboard-layout): add CurrentLayout and show it on the panel

JoshStrobl:
## Description

The applet asks the daemon to switch layouts and the daemon relays that to the labwc bridge, which writes XKB_DEFAULT_LAYOUT and reloads the compositor without going through locale1. Nothing reported back, so neither the applet nor the daemon could tell which of the configured layouts was in use.

The bridge now reads the layout back out of the environment file it writes, watching it with a file monitor so the value follows what landed on disk, and publishes it as CurrentLayout. The applet displays it, restoring the label the Vala applet had before the C port. Until the bridge publishes one, the applet falls back to the first configured input source, which is the layout the bridge builds from that same list at startup.

The reason I opted to implement it so the bridge notifies the daemon is that it is responsible for the environment file writing and compositor reloading (so it knows when that is done), and it keeps the daemon compositor-agnostic by pushing the responsibility for communicating the current layout / layout changed to the daemon for propagation throughout Budgie to anything listening, rather than the daemon reading environment files.

valac exports the property but never emits PropertiesChanged, so the daemon emits it by hand or client proxy caches go stale.

Also restores the panel_position_changed override the C port dropped, which matters again now that the button has a label beside the icon.

## Test plan

1. Test out the keyboard-layout applet PR
2. Be happy that layout is changing but sad that it isn't obvious because clicking on a layout in popover doesn't look like it's changing anything.
3. Check out this PR
4. Restart the budgie-daemon (`budgie-daemon --replace &`), kill the bridge (`killall -9 python3` or do something more surgical), start bridge (`/usr/bin/python3 /usr/libexec/budgie-desktop/labwc_bridge.py`), restart panel (`budgie-panel --replace`)
5. See layout in applet again
6. Open popover with layouts listed, change layout, see layout change
7. Be happy

### Submitter Checklist


EbonJaeger:
Just for completeness sake, does the current layout detection stuff still work when the layout is changed via outside means, such as `localectl`, or via DBus (`locale1`)?

JoshStrobl:
That I don't know, right now the bridge prioritizes the input-sources when it exists and has sources, before falling back to locale1.

I'll wipe my input sources and check.

JoshStrobl:
@EbonJaeger

Upon resetting input-sources, the applet crashed, likely because all the values from the gsettings fell out from under it. I'll see if the same happens when you add / delete single input sources, and if it's a segfault introduced in my code.

Regardless, I restarted the labwc bridge and then started up the panel, and the label shown on the applet was "us" and in the applet popover it reported "English (US)". Using localectl to change to Finnish (`localectl set-x11-keymap fi`) immediately propagated to the environment, which then propagated to CurrentLayout and updated the label and popover.

Changing it back to `us` updated the value then too.

JoshStrobl:
Okay adding / deleting input sources makes the applet unhappy. Will get that fixed.

EbonJaeger (approved):
LGTM, thanks!

PR #979, opened 2026-09-19
Remove GTK theme blacklist

City-busz:
Since commit 673231e658216ce72677985377c17ecf1ea839f3 the existence of the GTK3 CSS is explicitely checked, there is no need to add those themes to the blacklist which does not provide a GTK3 theme variant. The only themes that actually provide a GTK3 variant is Adwaita and Breeze, but these GTK themes work without issues, only the corresponding icon themes are problematic.


JoshStrobl:
No we're not removing the blacklist. This is the second time I'm having to reject this.

PR #980, opened 2026-09-19
Add option to disable sync with GtkSettings file

City-busz:
This allows the users to disable writing the cursor and theme settings in the user's local config settings.ini file. Saving into that file is not always desirable, because it's applied also for other sessions.


JoshStrobl:
No plans on changing the current behavior.

PR #981, opened 2026-09-19
Expose the color scheme setting

City-busz:
Instead of keeping the panel dark mode preference in sync with the global color scheme, add it as a separated option to Budgie Settings. This allows to select the desired dark mode preference for the applications independently from the panel.


JoshStrobl:
No plans on changing the current behavior.

PR #982, opened 2026-09-19
Save window buttons layout directly to gnome wm schema

City-busz:
This eliminates the need to keep the values in sync with the setting stored in Budgie GSettings schema, and also allows users to set custom buttonn ayouts for their needs.


JoshStrobl:
Been over this, we want more separation from gnome schemas, not less. Not accepting this.

